PR #1252, opened 2017-02-10
DPRO-3344: fix journal name bug in search results

sjtower:


sjtower:
Paired with @Fernando

PR #1254, opened 2017-02-15
DPRO-3346: Move theme tree structure from wombat.yaml

rskonnord-plos:
Note: will require a wombat.yaml update after merging.

### DIFF
--- a/src/main/java/org/ambraproject/wombat/config/YamlConfiguration.java
+++ b/src/main/java/org/ambraproject/wombat/config/YamlConfiguration.java
@@ -24,17 +24,15 @@ package org.ambraproject.wombat.config;
 
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.gson.GsonBuilder;
-import org.ambraproject.wombat.config.site.SiteSet;
-import org.ambraproject.wombat.config.theme.Theme;
-import org.ambraproject.wombat.config.theme.ThemeTree;
+import org.ambraproject.wombat.config.theme.FilesystemThemeSource;
 
+import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.Collection;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 
 /**
@@ -99,13 +97,10 @@ public class YamlConfiguration implements RuntimeConfiguration {
   }
 
   @Override
-  public ThemeTree getThemes(Collection<? extends Theme> internalThemes, Theme rootTheme) throws ThemeTree.ThemeConfigurationException {
-    return ThemeTree.parse(input.themes, internalThemes, rootTheme);
-  }
-
-  @Override
-  public SiteSet getSites(ThemeTree themeTree) {
-    return SiteSet.create(input.sites, themeTree);
+  public ImmutableList<FilesystemThemeSource> getThemeSources() {
+    return input.themeSources.stream()
+        .map(File::new).map(FilesystemThemeSource::new)
+        .collect(ImmutableList.toImmutableList());
   }
 
   private final CacheConfiguration cacheConfiguration = new CacheConfiguration() {
@@ -229,7 +224,7 @@ public class YamlConfiguration implements RuntimeConfiguration {
      * All such fields are immutable by convention. They should be set only by the YAML deserializer.
      * The intent of the @Deprecated annotation is to raise a warning in the IDE if a human refers
      * to them in code. (The reflective code in the library won't care at runtime of course.)
-
+     *
      * ---------------- Input fields (and boring boilerplate setters) are below this line ----------------
      */
 
@@ -239,8 +234,7 @@ public class YamlConfiguration implements RuntimeConfiguration {
     private String compiledAssetDir;
     private String rootPagePath;
     private List<String> enableDevFeatures;
-    private List<Map<String, ?>> themes;
-    private List<Map<String, ?>> sites;
+    private List<String> themeSources;
 
     private CacheConfigurationInput cache;
     private HttpConnectionPoolConfigurationInput httpConnectionPool;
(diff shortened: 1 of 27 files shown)


src/main/java/org/ambraproject/wombat/controller/FigurePageController.java
```diff
@@ -101,7 +102,7 @@ public String renderFigurePage(Model model, @SiteParam Site site,
 
     Map<String, ?> figureMetadata = articleMetadata.getFigureView().stream()
         .filter((Map<String, ?> fig) -> fig.get("doi").equals(assetPointer.getAssetDoi()))
-        .findAny().orElseThrow(RuntimeException::new);
+        .collect(MoreCollectors.onlyElement());
```
rskonnord-plos: This is unrelated, but I used `MoreCollectors.onlyElement()` elsewhere and figured that was as good an excuse as any to tighten up other places where we ought to use it.

Using `.findAny().orElseThrow(RuntimeException::new)` expects 1 element, throws an error on 0, and silently ignores extra elements in case there are more than 1. Using `.collect(MoreCollectors.onlyElement())` throws an error if there are fewer *or more* than 1.

src/main/java/org/ambraproject/wombat/controller/PowerPointController.java
```diff
@@ -119,7 +120,7 @@ public void download(HttpServletRequest request, HttpServletResponse response,
           String figureDoi = (String) figure.get("doi");
           return figureDoi.equals(assetPointer.getAssetDoi());
         })
-        .findAny();
+        .collect(MoreCollectors.toOptional());
```
rskonnord-plos: Similar to `MoreCollectors.onlyElement()`, `toOptional()` expects 0 or 1 elements but, unlike `findAny()`, will throw an exception if there is more than 1.

src/main/java/org/ambraproject/wombat/config/theme/ThemeBuilder.java
```diff
@@ -0,0 +1,115 @@
...
+ * An object that contains all the data necessary to construct a {@link Theme} object.
+ * <p>
+ * The purpose of this intermediate step is to allow a {@code Theme}'s parents to be constructed first. Because a {@code
+ * Theme} has immutable pointers to its parents, they must all be constructed in one pass while building a {@link
+ * ThemeTree}. Code that parses theme specifications can represent it as {@code ThemeBuilder} objects because a {@code
+ * ThemeBuilder} only needs to refer to parents by their key string, not as objects.
+ * <p>
+ * A {@link ConstructorFunction} object encapsulates whatever additional values are necessary to construct a theme of
+ * the particular subtype. For example, for a {@link FileTheme}, the {@code ConstructorFunction} would hold on to a
+ * pointer to a {@link java.io.File} object and pass it to the {@code FileTheme} constructor.
```
rskonnord-plos: Feedback requested: does this make sense?

I experimented with other ways of doing the same thing, mainly by making `ThemeBuilder` an abstract class with one extension for `FileTheme`, etc. This way wound up being much less verbose.

All this indirection is worth it so that we can separate out the theme-building step and put it closer to where we parse the configuration. This already matters for injecting themes into unit tests (see the updated tests in this PR). It may matter even more in the future if we ever want to plug in themes from somewhere other than the server's local filesystem.

src/main/java/org/ambraproject/wombat/config/theme/FilesystemThemeSource.java
```diff
@@ -0,0 +1,160 @@
...
+        return super.visitFile(file, attrs);
+      }
+    });
+    return hits;
+  }
+
+  private static ThemeBuilder.ConstructorFunction<FileTheme> createConstructorFunction(File themeFile) {
+    File themeDirectory = Objects.requireNonNull(themeFile.getParentFile());
+    return (key, parentObjects) -> new FileTheme(key, parentObjects, themeDirectory);
+  }
```
rskonnord-plos: Regarding https://github.com/PLOS/wombat/pull/1254#discussion_r101414286, this is a practical example of how a `ConstructorFunction` plugs into a `ThemeBuilder`. Here, it seems pretty straightforward IMHO.

The one awkward part is how the key is passed to the `ThemeBuilder` constructor, which then passes it back to this `ConstructorFunction`. Tucking the lambda away in this `createConstructorFunction` method actually sweeps this blemish under the rug pretty nicely -- if you were to inline `createConstructorFunction` into `parseThemeDescription`, you can see how `key` would become overloaded. I experimented with a few different ways to move the key value around and this one seemed the least brittle. I like doing it this way because we create the `ConstructorFunction` seeing only the `File` -- i.e., the thing that this `FileTheme` uniquely depends on that wouldn't apply to other `Theme` subclasses.

src/main/java/org/ambraproject/wombat/config/YamlConfiguration.java
```diff
@@ -98,14 +99,25 @@ public String getRootPagePath() {
...
 
-  @Override
-  public ThemeTree getThemes(Collection<? extends Theme> internalThemes, Theme rootTheme) throws ThemeTree.ThemeConfigurationException {
-    return ThemeTree.parse(input.themes, internalThemes, rootTheme);
+  private static ThemeSource<?> parseThemeSource(Map<String, ?> map) {
+    String type = (String) map.get("type");
+    if (type == null) throw new RuntimeException("Theme source must have type");
+    // Type is a hook for other ThemeSource types that may exist in the future.
+    // Currently the only supported value is "filesystem".
+    if (type.equals("filesystem")) {
```
sjtower: Slightly confused about what 'type' means here. What is the use case for other types, and can we add them to an enum or something?

rskonnord-plos: The main anticipated use case is some sort of remote source or service, something we access by URL.

An enum sounds fine; will add.

src/main/java/org/ambraproject/wombat/config/theme/FilesystemThemeSource.java
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * A theme source consisting of a directory on the server's local file system.
+ * <p>
+ * A file named {@code sites.yaml}, at the directory's root level, defines the sites. Any file named {@code theme.yaml}
+ * that appears in a subdirectory defines that subdirectory as a theme; the content of the YAML file indicates the
+ * theme's key and its parents.
+ */
+public final class FilesystemThemeSource implements ThemeSource<FileTheme> {
+
+  private final File root;
```
sjtower: Name could be more descriptive, EG rootYamlFile

rskonnord-plos: It's actually the root directory, so yeah. :wink:

rskonnord-plos:
I'm not going to check it into this branch because of the large amount of noise it would create, but, since we're messing with the `ThemeTree` class already, I think now would be a great time to rename it to `ThemeGraph` like I've wanted to forever. (See the class-level Javadoc for explanation.)

PR #1257, opened 2017-02-28
DPRO-3287: add plos-themes as deb dependency

sjtower:
Paired with @johnfesenko

### DIFF
--- a/src/deb/control/control
+++ b/src/deb/control/control
@@ -3,6 +3,6 @@ Version: [[version]]
 Section: java
 Priority: optional
 Architecture: all
-Depends: default-jre | java8-runtime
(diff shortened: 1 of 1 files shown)


src/deb/control/control
```diff
@@ -3,6 +3,6 @@ Version: [[version]]
 Section: java
 Priority: optional
 Architecture: all
-Depends: default-jre | java8-runtime
+Depends: default-jre | java8-runtime | plos-themes
```
rskonnord-plos: Is this unavoidable? I know the Debian packaging is already a PLOS-only thing. But I hate to see `plos-themes` mentioned in Wombat as a general pattern, especially if it would be one more thing to clean up if we ever generalize these Debian packages for public use.

PR #1274, opened 2017-04-24
DPRO-3420 Removing duplicated code

sbassi:
Eliminating duplicated code. This should be deployed with dev-DPRO-3420 in plos-themes.


src/main/webapp/WEB-INF/themes/desktop/ftl/common/head.ftl
```diff
@@ -76,6 +76,8 @@
 <#include "baseJs.ftl" />
 <#include "additionalHeadTracking.ftl" />
 
+<script type="text/javascript" src="http://www.google.com/recaptcha/api/js/recaptcha_ajax.js"></script>
```
anilop: I don't think we should have this in the head tag for all pages and definitely not in wombat.  Since we don't use captcha in all the pages.

sbassi:
@anilop deleted that call.

PR #1278, opened 2017-05-15
RARO-53 Display articles from the preprints bucket

msingh05:


src/main/java/org/ambraproject/wombat/service/remote/AbstractContentApi.java
```diff
@@ -128,7 +128,7 @@ private URI buildUri(ContentKey key, RequestMode mode) throws IOException {
     RepoConfig repoConfig = getRepoConfig();
     UrlParamBuilder requestParams = UrlParamBuilder.params();
     key.setParameters(requestParams);
-    String repoBucketName = repoConfig.bucketName;
+    String repoBucketName = key.getBucketName() != null ? key.getBucketName() : repoConfig.bucketName;
```
sjtower: The change in rhino guarantees that a bucket name will be included with the article metadata, correct? So please remove the repoConfig.bucketName code throughout this class.

msingh05: I tried that first, but there are other places in wombat where contentKey is created, but bucketName is not available. So contentKey may not always have a bucketName I think. If I change everywhere it may break other unintended pieces, or will require significantly more effort to fix those places.

sjtower: :+1:

PR #1281, opened 2017-05-23
RARO-88: break out solr configuration

sjtower:
@msingh05


src/test/java/org/ambraproject/wombat/config/TestRuntimeConfiguration.java
```diff
@@ -46,8 +46,8 @@ public URL getServer() {
   }
 
   @Override
-  public Optional<URL> getSolrServer() {
-    return Optional.empty();
+  public Optional<SolrConfiguration> getSolrConfiguration() {
+    return null;
```
msingh05: Why not Optional.empty()?

sjtower: fixed

PR #1286, opened 2017-07-06
RARO-299: fail gracefully if solr is down (article page)

sjtower:


src/main/java/org/ambraproject/wombat/controller/ArticleController.java
```diff
@@ -776,8 +777,9 @@ private XmlContent getXmlContent(Site site, ArticlePointer articlePointer,
             String citationJournalKey;
             try {
               citationJournalKey = doiToJournalResolutionService.getJournalKeyFromDoi(doi, site);
-            } catch (SolrUndefinedException e) {
+            } catch (SolrUndefinedException | EntityNotFoundException e) {
```
anilop: Do we want to log these two exceptions separately?

msingh05: I am ok with it. I just want to know which app is throwing error and we can figure out why solr is not good in this case.

PR #1288, opened 2017-07-19
RARO-324: remove captcha from media curation form

sjtower:
I've already created v3.3.5rc which contains these changes.


anilop:
deleting because these changes are already in v3.3.5rc

PR #1291, opened 2017-08-03
RARO-312: handle case of no volume, issue for article metadata displa…

sjtower:
…y. ppub and epreprint pub-date tags will now display properly

### DIFF
--- a/src/main/webapp/WEB-INF/themes/desktop/xform/article-transform.xsl
+++ b/src/main/webapp/WEB-INF/themes/desktop/xform/article-transform.xsl
@@ -205,19 +205,25 @@
           </xsl:choose>
         </xsl:for-each>
         <!-- pub year -->
-        <xsl:text> (</xsl:text>
-        <xsl:choose>
-          <xsl:when test="pub-date[@pub-type='collection']/year | pub-date[@pub-type='ppub']/year">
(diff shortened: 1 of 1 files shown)


src/main/webapp/WEB-INF/themes/desktop/xform/article-transform.xsl
```diff
@@ -206,7 +206,17 @@
...
         <xsl:text> (</xsl:text>
-        <xsl:value-of select="pub-date[@pub-type='collection']/year | pub-date[@pub-type='ppub']/year"/>
+        <xsl:choose>
+          <xsl:when test="pub-date[@pub-type='collection']/year | pub-date[@pub-type='ppub']/year">
+            <xsl:value-of select="pub-date[@pub-type='collection']/year | pub-date[@pub-type='ppub']/year"/>
+          </xsl:when>
+          <xsl:when test="pub-date[@pub-type='epub']/year">
+            <xsl:value-of select="pub-date[@pub-type='epub']/year"/>
+          </xsl:when>
+          <xsl:otherwise>
```
anilop: this is a strange otherwise, I think we should test for it. and include nothing - no parenthesis when we have no "pubdate".

sjtower:
Implemented code review changes, please re-review

PR #1299, opened 2017-08-31
Dev raro 449

msingh05:


src/main/webapp/WEB-INF/themes/desktop/sass/ApertaRxiv/_article_page.scss
```diff
@@ -112,10 +112,52 @@ main article {
...
         margin-right: 2ch;
+        position: relative;
+      }
+
+    }
+  }
+
+  .share-options {
+    opacity: 1;
+    z-index: 10;
```
anilop: Do we need this? kube seems to add a z-index - but I may be wrong. if not we should do this map thing instead. It is super handy, this way instead of rooting through the code trying to figure out which z-index is best you can just name it.

you create a map - we could put it in components.scss
```
$zindex: (
    modal : 90,
    overlay : 80,
    dropdown : 68,
    invisible : -1,
);
```

then you can call it  this:
`  z-index: map-get($zindex, modal);`

- we can use less options

we do it in the code right now in utlities.scss

src/main/webapp/WEB-INF/themes/desktop/resource/js/pages/aperta_article.js
```diff
@@ -63,4 +63,9 @@ $(document).ready(function () {
         $(item).append(a);
     });
 
+    $("#share-options li a").click(function(event) {
+        // TODO: link is not automatically opened, don't know why. So using window.open.
+        console.log("clicked " + event.currentTarget.getAttribute("href"));
```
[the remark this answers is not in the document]
msingh05: done

src/main/webapp/WEB-INF/themes/desktop/sass/ApertaRxiv/_article_page.scss
msingh05: it didn't automatically add z-index. but after i moved .share-options to end of the file, the article text does not overlap on top of that anymore. i remove z-index and opacity.

PR #1300, opened 2017-09-05
RARO-364: Wombat backend prep for anonymous comments

sjtower:


src/main/java/org/ambraproject/wombat/controller/ArticleController.java
```diff
@@ -195,6 +195,10 @@ private void addCommentAvailability(Model model) {
     model.addAttribute("areCommentsDisabled", runtimeConfiguration.areCommentsDisabled());
   }
 
+  private boolean isPreprints(Site site) {
+    return site.getType() != null && site.getType().equals("preprints");
+  }
+
```
msingh05: Should this method be moved to somewhere as public so that it can be called from other places, instead of comparing with "preprints" at every place?

sjtower: I will simply add it as a method on the Site class.

PR #1308, opened 2017-09-15
RARO-403: revert kube.js

sjtower:
@anilop


anilop:
This was just a test no need to merge

PR #1309, opened 2017-09-19
RARO-505: paginate LOCKSS pages

sjtower:


src/main/java/org/ambraproject/wombat/controller/LockssController.java
```diff
@@ -97,14 +100,29 @@ public String getMonthsForYear(@SiteParam Site site, @PathVariable String year,
...
     model.addAttribute("year", year);
     model.addAttribute("searchResult", searchResult);
+    model.addAttribute("nextCursorMark", searchResult.get("nextCursorMark"));
+
+    int pageNumberCount = Integer.parseInt(pageNumber);
+    final int listStartNumber = pageNumberCount * MAXIMUM_SOLR_RESULTS + 1;
+    model.addAttribute("listStart", listStartNumber);
+
+    final boolean isLastPage = (Double) searchResult.get("numFound")
+        <= listStartNumber + MAXIMUM_SOLR_RESULTS;
```
msingh05: I think line 121 should change <= to < for comparison.

suppose max-solr-results = 10, and numFound = 61, so results will show in 7 pages. when pageNumber=5 (it is index starting with 0), then listStartNumber = 5*10+1=51 (it is index starting with 1), then isLastPage = (61 <= 51+10) = true, but there is one more page after this one.

sjtower: nice catch! Pushed the change.

PR #1310, opened 2017-09-26
RARO-530: consolidate configs as yaml

sjtower:


anilop:
We should create empty versions of almConfigJs.ftl  and such files in wombat. That's what we usually do. It supports break the generic case AND it makes it clear what is happening.

PR #1311, opened 2017-09-27
RARO-342, comment functionality

msingh05:


src/main/webapp/WEB-INF/themes/desktop/sass/ApertaRxiv/components/_component.scss
```diff
@@ -77,6 +77,126 @@
...
+        }
+      }
+      &:not(.active) {
+        margin-bottom: 4px;
+      }
+    }
+
+    .arrow {
+      display: inline-block;
+      font-size: 18px;
```
anilop: We should be using variables for these

PR #1315, opened 2017-10-02
RARO-515 as per review

msingh05:


src/main/webapp/WEB-INF/themes/desktop/sass/ApertaRxiv/_article_page.scss
```diff
@@ -158,7 +159,7 @@ main article {
 
         }
         &:hover {
-          background-color: $color-button-secondary;
+          background-color: #7b9abf;
```
anilop: Oh Man, another color?

PR #1316, opened 2017-10-02
RARO-526: the "Search icon" is not displayed along with the "X" on mo…

msingh05:
…bile


src/main/webapp/WEB-INF/themes/desktop/sass/ApertaRxiv/components/_icons.scss
```diff
@@ -42,7 +42,10 @@
   @if ($responsive: true) {
     @include breakpoint(sm) {
       background-size: $sml-size;
+      width: 32px;
```
anilop: WE should make these variables. since they are icons that might be used elsewhere.

PR #1318, opened 2017-10-05
RARO-565 Design feedback - Fixes Authors - Additional fixes

msingh05:


src/main/webapp/WEB-INF/themes/desktop/sass/ApertaRxiv/components/_component.scss
```diff
@@ -223,6 +224,8 @@
     }
     &.active {
       font-weight: bold;
+      zindex: map-get($zindex, topvisible);
+
```
msingh05: @anilop did change. Also as we talked about margins, I am handling them in current metrics ticket. so this ticket can me merged.

PR #1322, opened 2017-10-12
Increased search client timeout.

randy-tano:


randy-tano:
Can't use as branch was cut from master.

PR #1343, opened 2017-11-15
RARO-555: get realtime counter data for metrics tab

sjtower:


anilop:
How are we getting the monthly breakdowns for the graph here? Those come from a different counter endpoint?

sjtower:
Monthly breakdown data hasn't changed. It uses the cumulative totals, which we've changed to get from counter instead of ALM. The code is a bit confusing but I don't see any explicit calls to ALM for the monthly data.

anilop:
What we get back from ALM is a complete breakdown of data. There are two endpoints for the new counter. 

One is totals: http://counter-201.soma.plos.org/api/v1.0/stats/totals/doi/10.1371/journal.pone.0043202 
 the other is a month by month breakdown:
 http://counter-201.soma.plos.org/api/v1.0/stats/
doi/10.1371/journal.pone.0043202. 

The totals work for the table but not for the graph below. You were saying we get the historical month by month data for free and I"m making sure this assumption is true. Which endpoint are we using?

sjtower:
We're using the former endpoint that is simply the totals.

I didn't say that we get historical month-by-month data for free. What I said was that the table data depends on the graph data, which is true afaik by looking at the code. The month-by-month details appear to be calculated off of the cumulative total, only. But I'm having a hard time grokking the graph code. lets sit down and look at it together.

PR #1344, opened 2017-11-16
RARO-694: Implement honeypot for commenting on ApertaARxiv

msingh05:


src/main/java/org/ambraproject/wombat/controller/ArticleController.java
```diff
@@ -313,12 +313,21 @@ public Object receiveNewComment(HttpServletRequest request,
...
+                                  @RequestParam(value = "authorAffiliation", required=false) String authorAffiliation,
                                   @RequestParam(value = "ciStatement", required = false) String ciStatement,
                                   @RequestParam(value = "target", required = false) String parentArticleDoi,
                                   @RequestParam(value = "inReplyTo", required = false) String parentCommentUri,
                                   @RequestParam(value = RECAPTCHA_CHALLENGE_FIELD, required=false) String captchaChallenge,
                                   @RequestParam(value = RECAPTCHA_RESPONSE_FIELD, required=false) String captchaResponse)
       throws IOException {
+
+    // honeypot for bot
+    if (authorPhone != null && !authorPhone.isEmpty() || authorAffiliation != null && !authorAffiliation.isEmpty()) {
```
randy-tano: I wonder if checking that `authorPhone` or `authorAffiliation` was passed is a good test.  That is, we're expecting those fields, and so, could never be null.  If those fields are missing then we have an invalid submission, which could mean that it's a bot .  What do you think?

sjtower: Please include a detailed comment explaining the honeypot and how it works as well

msingh05: The authorPhone and authorAffiliation are not required for this API. I added those in the form, and made them hidden in CSS. So if some non-empty value is passed for any of those fields it means it is not a regular form submission. So mark it as a bot, and don't proceed. But don't alarm the bot, hence return success.

anilop: so instead of adding a meaningless field we add something that seems meaningful? I agree, let's add more comments so we don't go in circles on it for some reason.

PR #1356, opened 2018-01-05
RARO-54 RSS feeds for comments are broken

msingh05:


src/main/java/org/ambraproject/wombat/service/CommentServiceImpl.java
```diff
@@ -128,6 +145,28 @@ private void addCreatorData(Collection<Map<String, Object>> rootComments) {
     }));
   }
 
+  private void addArticleTitle(Collection<Map<String, Object>> rootComments) {
```
randy-tano: Since this is `private` and it never uses any instance variables, you can probably make it `static` too.

src/main/java/org/ambraproject/wombat/service/CommentServiceImpl.java
```diff
@@ -128,6 +145,28 @@ private void addCreatorData(Collection<Map<String, Object>> rootComments) {
...
+    // Gather up all distinct article DOIs in the forest
+    Set<String> articleDois = Collections.synchronizedSet(new HashSet<>());
```
randy-tano: You may get better performance by using `java.util.concurrent.ConcurrentHashMap`, rather than the **synchronized** version, as the **concurrent** version will avoid locking the entire set.
```
Set<String> articleDois = ConcurrentHashMap.newKeySet();
```

src/main/java/org/ambraproject/wombat/service/CommentServiceImpl.java
msingh05: Cannot be made static as it uses to this::requestArticleTitle

src/main/java/org/ambraproject/wombat/service/CommentServiceImpl.java
msingh05: changed.

PR #1359, opened 2018-02-02
Use / as the path when starting wombat with mvn tomcat6:run

egh:
The root (`/`) context is used by rhino and wombat when starting with the tomcat maven plugin. This changes wombat to use it also.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -364,6 +364,7 @@
         <artifactId>tomcat6-maven-plugin</artifactId>
         <version>2.1</version>
         <configuration>
+          <path>/</path>
           <systemProperties>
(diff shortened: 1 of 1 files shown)


sjtower:
LGTM, although we run Wombat in Tomcat 7:

`tomcat:run -Dmaven.tomcat.port=8234 -Dwombat.configDir=/etc/ambra`

sjtower:
This makes me wonder why we have the tomcat6 plugin in the pom at all...

egh:
@sjtower Thanks! wombat, rhino and crepo all use the tomcat6 maven plugin. I guess it should be upgraded

PR #1361, opened 2018-02-03
Upgrade to tomcat7 maven plugin

egh:
We use tomcat 7 in production, we should upgrade the maven plugin used for testing.


sgatchell:
👍

PR #1369, opened 2018-02-07
RARO-945

msingh05:
Fix view article link references

### DIFF
--- a/src/main/java/org/ambraproject/wombat/service/ParseReferenceService.java
+++ b/src/main/java/org/ambraproject/wombat/service/ParseReferenceService.java
@@ -124,12 +124,10 @@ public class ParseReferenceService {
         boolean useDoi = false;
         Node previousNode = extLink.getPreviousSibling();
         if (previousNode != null) {
-          String previousText = extLink.getPreviousSibling().getTextContent();
+          String previousText = previousNode.getTextContent();
           if (!Strings.isNullOrEmpty(previousText)) {
             previousText = previousText.trim();
-            if (previousText.equalsIgnoreCase("doi:") || previousText.equalsIgnoreCase("doi")) {
-              useDoi = true;
-            }
+            useDoi = previousText.equalsIgnoreCase("doi:") || previousText.equalsIgnoreCase("doi");
           }
         }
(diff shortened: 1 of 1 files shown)


anilop:
I'm concerned this will only solve the case of the link with doi in the url.

What about the following case from the XML:
`<comment>Available from: <ext-link ext-link-type="uri" xlink:type="simple" xlink:href="http://www.sciencemag.org/content/333/6051/1878.abstract">http://www.sciencemag.org/content/333/6051/1878.abstract</ext-link></comment>` 

outputs as:
`https://doi.org/http://www.sciencemag.org/content/333/6051/1878.abstract`

msingh05:
This should work. Lets see if QA can find more use cases.

src/main/java/org/ambraproject/wombat/service/ParseReferenceService.java
```diff
@@ -120,11 +120,25 @@ public String getValue() {
       NodeList extLinkList = element.getElementsByTagName("ext-link");
       if (!ParseXmlUtil.isNullOrEmpty(extLinkList)) {
         Element extLink = (Element) extLinkList.item(0);
+        // use the ext-link as doi, only if ext-link previous text is "doi:" or "doi"
+        boolean useDoi = false;
+        Node previousNode = extLink.getPreviousSibling();
+        if (previousNode != null) {
+          String previousText = extLink.getPreviousSibling().getTextContent();
```
randy-tano: I suggest using `previousNode`, since it's already assigned as `extLink.getPreviousSibling()`.

src/main/java/org/ambraproject/wombat/service/ParseReferenceService.java
```diff
@@ -120,11 +120,25 @@ public String getValue() {
...
+          if (!Strings.isNullOrEmpty(previousText)) {
+            previousText = previousText.trim();
+            if (previousText.equalsIgnoreCase("doi:") || previousText.equalsIgnoreCase("doi")) {
```
randy-tano: Would this work too?
```
useDoi = previousText.equalsIgnoreCase("doi:") || previousText.equalsIgnoreCase("doi");
```

PR #1374, opened 2018-02-13
Use default $HOME/ambra/config for wombat.configDir

egh:
~/ambra/config is used in the open source docs as the location for configuration. Why not make it the default when a user runs `mvn tomcat6:run`?


sgatchell:
"just works" 👍

PR #1376, opened 2018-02-15
Dev raro 853

msingh05:


src/main/webapp/WEB-INF/themes/desktop/xform/article-transform.xsl
```diff
@@ -1046,26 +1046,19 @@
...
+                          <xsl:value-of select="concat('http://www.ncbi.nlm.nih.gov/pubmed/', $pmid)"/>
+                        </xsl:attribute>
+                        <xsl:attribute name="target">_new</xsl:attribute>
+                        <xsl:attribute name="title">Go to article in PubMed</xsl:attribute>
+                        PubMed/NCBI
+                      </xsl:element>
                     </xsl:element>
-                  </xsl:element>
+                  </xsl:if>
                   <xsl:element name="li">
```
msingh05: not sure why above code is also showing up. It was from another ticket.

src/main/webapp/WEB-INF/themes/desktop/xform/article-transform.xsl
```diff
@@ -2260,7 +2253,6 @@
 
   <!-- 1/4/12: Ambra-specific template -->
   <xsl:template name="citationComment">
-    <xsl:apply-templates />
```
msingh05: this is the only change.

PR #1377, opened 2018-02-15
AMBR-62: add counter to mobile metrics tab

sjtower:


anilop:
This should be against a hotfix branch created off the RC

PR #1388, opened 2018-03-01
AMBR-63

msingh05:
Display DOIs for non-journal references too


src/main/webapp/WEB-INF/themes/desktop/xform/article-transform.xsl
```diff
@@ -2492,12 +2492,18 @@
 
 
     <xsl:choose>
-      <xsl:when test="not(ancestor::ref-list) or not(substring($previousText, string-length($previousText)-3)='doi:')">
+      <xsl:when test="not(ancestor::ref-list) or ancestor::*/@publication-type='journal' and not(substring($previousText, string-length($previousText)-3)='doi:')">
```
randy-tano: :+1:

PR #1389, opened 2018-03-08
AMBR-3 debian: do not db_purge in postinst

egh:
debian packages should not clear the questions answered when installing. This
should happen when the package is purged.


egh:
This is a fix to the problem with deploying our debian packages. @anilop @sjtower

wryfi:
:thumbsup:

PR #1392, opened 2018-03-13
AMBR-213, AMBR-215: replace captcha with honeypot

sjtower:


src/main/java/org/ambraproject/wombat/controller/ArticleController.java
```diff
@@ -327,19 +324,10 @@ public Object receiveNewComment(HttpServletRequest request,
...
-                                  @RequestParam(value = RECAPTCHA_RESPONSE_FIELD, required = false) String captchaResponse)
+                                  @RequestParam(value = "inReplyTo", required = false) String parentCommentUri)
       throws IOException {
 
-    // honeypot for bot.
-    // authorPhone and authorAffiliation are fake parameters, which are present in the form, but hidden via CSS.
-    // A bot will likely fill one or both of these. But a human will not see those fields to fill them.
-    // If any of these parameters are non-empty, then mark it as a bot, and do not proceed further.
-    // However, return success response to avoid alarming the bot.
-
```
msingh05: we can leave documentation as it is.

PR #1405, opened 2018-04-03
AMBR-29: fix font height in publish menu button

sjtower:


anilop:
THis should not be in vendor code. we shoudl have changed the mega-menu css

PR #1408, opened 2018-04-10
AMBR-29: Submit now button on Plos mega-menu doesn't appear correctly

msingh05:


src/main/webapp/WEB-INF/themes/desktop/sass/sections/_header.scss
```diff
@@ -379,6 +379,8 @@ $width-nav: $block-grid-elements - $width-logo;
       a {
         padding-top: 0;
         padding-bottom: 0;
+        vertical-align: middle;
+        line-height: 28px;
```
anilop: should we make this rem-calc(28)? to be consistent

anilop: Also:    align-items: center;
    display: inline-flex;

will make it possible to be more flexible - no  pun intended and not have to set the height

anilop:
LGTM

PR #1413, opened 2018-04-18
SERV-61: Capture consent on feedback form - AMBRA

msingh05:


src/main/java/org/ambraproject/wombat/controller/FeedbackController.java
```diff
@@ -89,6 +89,7 @@ public String receiveFeedback(HttpServletRequest request, HttpServletResponse re
                                 @RequestParam("subject") String subject,
                                 @RequestParam("name") String name,
                                 @RequestParam("userId") String userId,
+                                @RequestParam(value = "consent", required = false) String consent,
```
randy-tano: Should this be `required`?  Since, the validation doesn't allow for a `null` value.

msingh05: When consent checkbox is not checked, the browser does not send "consent" parameter. So it gives 404 Not Found if I make it required. That's why I had to make it optional, and then check if it is null, which means browser did not send the parameter.

randy-tano: Got it.

PR #1436, opened 2018-05-30
AMBR-411: Optimize parseArticleReferences method while rendering arti…

msingh05:
…cles

### DIFF
--- a/src/main/java/org/ambraproject/wombat/service/ParseXmlServiceImpl.java
+++ b/src/main/java/org/ambraproject/wombat/service/ParseXmlServiceImpl.java
@@ -41,6 +41,9 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.stream.Collectors;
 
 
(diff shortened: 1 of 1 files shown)


msingh05:
Added code for using thread pool to run buildReferences (which internally calls solr query) in parallel. Tried both newCachedThreadPool and newFixedThreadPool(10). The first one has a risk of having too many thread at a time, if there are too many references or simultaneous article accesses. The second one reduces the time taken by at most a factor of 10. In my instrumented code, both gave similar results with a single article journal.ppat.1004272. Without the changes, building all the references takes 3+ seconds. With the changes building all references is less than 0.5 seconds. However the first run of the changed code takes longer by ~3seconds - I think because it creates threads on first run.

randy-tano:
You'll need to factor in synchronization issues with using an `ArrayList` (this [line](https://github.com/PLOS/wombat/blob/f0dc9590cd8ccfae18326c041b6e2ea9e3b5c2d0/src/main/java/org/ambraproject/wombat/service/ParseXmlServiceImpl.java#L87)).
```
List<Reference> references = new ArrayList<>();
```
**From the docs**:  Note that this implementation is not synchronized. If multiple threads access an ArrayList instance concurrently, and at least one of the threads modifies the list structurally, it must be synchronized externally. (A structural modification is any operation that adds or deletes one or more elements, or explicitly resizes the backing array; merely setting the value of an element is not a structural modification.)

PR #1439, opened 2018-06-05
AMBR-411: Optimize parseArticleReferences method while rendering arti…

msingh05:
…cles

### DIFF
--- a/src/main/java/org/ambraproject/wombat/controller/ArticleController.java
+++ b/src/main/java/org/ambraproject/wombat/controller/ArticleController.java
@@ -113,6 +113,7 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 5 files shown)


src/main/java/org/ambraproject/wombat/controller/ArticleController.java
```diff
@@ -806,15 +807,55 @@ private XmlContent getXmlContent(Site site, ArticlePointer articlePointer,
...
       StringWriter articleHtml = new StringWriter(XFORM_BUFFER_SIZE);
       try (OutputStream outputStream = new WriterOutputStream(articleHtml, charset)) {
         articleTransformService.transformArticle(site, articlePointer, references,
             new ByteArrayInputStream(xml), outputStream);
       }
 
+      long time5 = System.currentTimeMillis();
+      System.out.println("total time=" + (time5 - time1) + " transform=" + (time5 - time4)
+          + " references=" + (time4 - time1) + " including solr=" + (time3 - time2));
+
```
msingh05: I would like to keep these println as easier for testing. Once testing done, I can remove that. The last solr time is what is being optimized here. Previously, depending on the number and type of references it could be few seconds. Now it should be less than a second, preferably less than 200 millisecond. I tested with article journal.pone.0127757

src/main/java/org/ambraproject/wombat/service/DoiToJournalResolutionService.java
```diff
@@ -52,6 +57,37 @@ public String getJournalKeyFromDoi(String doi, Site site) throws IOException {
     }
     return journalKey;
   }
+
+  public List<String> getJournalKeysFromDois(List<String> dois, Site site) throws IOException {
+    if (dois.size() == 0) {
```
randy-tano: nit:  Can replace `dois.size() == 0` with `dois.isEmpty()`.

PR #1442, opened 2018-06-12
AMBR-411: bugfix

msingh05:


src/main/java/org/ambraproject/wombat/service/DoiToJournalResolutionService.java
```diff
@@ -77,15 +78,18 @@ public String getJournalKeyFromDoi(String doi, Site site) throws IOException {
...
-    if (docs.size() == dois.size()) {
-      journalKeys = docs.stream().map(data -> {
-          return (String) data.get("journal_key");
-      }).collect(Collectors.toList());
-    } else {
-      log.warn("incorrect solr result count " + docs.size() + " != " + dois.size()
-          + " for query=" + explicitDoiSearchQuery.getQuery());
-    }
+    Map<String, String> keysMap = new HashMap<String,String>();
+    docs.stream().forEach(data -> {
```
randy-tano: You can just call the `forEach` on the `docs` instance directly, rather than creating a **stream()** first.  For example:
```
docs.forEach(...)
```

src/main/java/org/ambraproject/wombat/controller/ArticleController.java
```diff
@@ -834,9 +834,13 @@ private XmlContent getXmlContent(Site site, ArticlePointer articlePointer,
         Reference ref = itRef.next();
         if (ref.getDoi() != null && ref.getDoi().startsWith("10.1371/")) {
           String key = itKey.next();
-          Reference.Builder builder = new Reference.Builder(ref);
-          Reference refWithLink = builder.setFullArticleLink(getLinkText(site, request, ref.getDoi(), key)).build();
-          referencesWithLinks.add(refWithLink);
+          if (!Strings.isNullOrEmpty(key)) {
```
randy-tano: nit: I usually avoid the **not:** **not** logic.  You can rewrite the `if`:
```
if (Strings.isNullOrEmpty(key)) {
    referencesWithLinks.add(ref);
} else {
    Reference.Builder builder = new Reference.Builder(ref);
    Reference refWithLink = builder.setFullArticleLink(getLinkText(site, request, ref.getDoi(), key)).build();
    referencesWithLinks.add(refWithLink);
}
```
Just a suggestion :)

msingh05:
done

PR #1444, opened 2018-06-19
AMBR-439 : Explicitly HTML escaped the rendered stacktrace.

randy-tano:
This PR explicitly HTML escapes the rendered stacktrace.  However, it looks like the template to render the wombat error page is in [plos-themes](https://github.com/PLOS/plos-themes/blob/develop/code/desktop/all_sites/ftl/error/error.ftl).   So, there's a correspond [PR](https://github.com/PLOS/plos-themes/pull/897) in **plos-themes** to fix the XSS issue.


anilop:
LGTM

PR #1449, opened 2018-07-13
AMBR-274 Reading a new attribut eform wombat.yaml called 'environment'

tpolischuk:
This is used to conditionally show the stacktrace only in production.

### DIFF
--- a/src/main/java/org/ambraproject/wombat/config/YamlConfiguration.java
+++ b/src/main/java/org/ambraproject/wombat/config/YamlConfiguration.java
@@ -98,6 +98,11 @@ public class YamlConfiguration implements RuntimeConfiguration {
     return input.rootPagePath;
   }
 
+  @Override
+  public String getEnvironment() {
+    return input.environment;
+  }
+
   /**
    * Future-proofing against the need for other ThemeSource types that may exist in the future (mainly, one that reads
    * from a remote source, probably by URL). Currently, the only supported type reads from the local filesystem.
@@ -334,6 +339,7 @@ public class YamlConfiguration implements RuntimeConfiguration {
     private String mailServer;
     private String compiledAssetDir;
     private String rootPagePath;
+    private String environment;
     private List<String> enableDevFeatures;
     private List<Map<String, ?>> themeSources;
 
@@ -376,6 +382,12 @@ public class YamlConfiguration implements RuntimeConfiguration {
     public void setRootPagePath(String rootPagePath) {
       this.rootPagePath = rootPagePath;
     }
+
+    /**
+     * @deprecated For access by reflective deserializer only
+     */
+    @Deprecated
+    public void setEnvironment(String environment) { this.environment = environment; }
 
     /**
      * @deprecated For access by reflective deserializer only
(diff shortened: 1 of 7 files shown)


src/main/java/org/ambraproject/wombat/config/YamlConfiguration.java
```diff
@@ -377,6 +383,12 @@ public void setRootPagePath(String rootPagePath) {
       this.rootPagePath = rootPagePath;
     }
 
+    /**
+     * @deprecated For access by reflective deserializer only
+     */
+    @Deprecated
+    public void setEnvironment(String environment) { this.environment = environment; }
```
sgatchell: I should have formatted this like the rest of them, rather than as a one liner.

src/main/webapp/WEB-INF/themes/root/app/approot.ftl
```diff
@@ -53,6 +53,7 @@
   Available under the
   <a href="http://creativecommons.org/licenses/by/2.0/deed.en">Creative Commons Attribution 2.0 Generic License.</a>
 </div>
+<h1>environment is: ${environment}</h1>
```
sgatchell: This was just a POC. If you want to remove it also revert AppRootPage.java.  If you want to keep it, an `<h1>` is a little too loud IMO.

src/main/webapp/WEB-INF/themes/root/app/error.ftl
```diff
@@ -41,10 +41,12 @@
 
     <p>Thank you for your patience.</p>
 
-    <div id="stackTrace">
-      <h2>Technical Information for Developers</h2>
-      <pre>${stackTrace?html}</pre>
-    </div>
+
+    <#if environment != "prod">
```
sgatchell: @anilop I see stackTrace is also presented in these files.  I assume that they're all in use and should get the same treatment? 

src/main/webapp/WEB-INF/themes/desktop/ftl/error/error.ftl 
src/main/webapp/WEB-INF/themes/mobile/ftl/error/error.ftl
src/main/webapp/WEB-INF/themes/root/ftl/error/error.ftl

Also, I see a reference in:
plos-themes/code/desktop/all_sites/ftl/error/error.ftl

sgatchell: Never mind about the reference in plos-themes, @tpolischuk covered it here: https://github.com/PLOS/plos-themes/pull/906

anilop: @sgatchell - Agreed, they should get the same treatment.

tpolischuk:
All Related PRs: 
https://github.com/PLOS-Formulas/wombat-formula/pull/80
https://github.com/PLOS/wombat/pull/1449
https://github.com/PLOS/plos-themes/pull/906

src/main/java/org/ambraproject/wombat/controller/AppRootPage.java
```diff
@@ -113,6 +113,7 @@ ModelAndView serveAppRoot() {
 
   private ModelAndView serveDebugPage() {
     ModelAndView mav = new ModelAndView("//approot");
+    mav.addObject("environment", runtimeConfiguration.getEnvironment());
```
sgatchell: Having removed the POC from `approot.ftl`, do we still need this?

PR #1452, opened 2018-07-18
ITI-3466 added maxdays with value of 5 in server.xml

evarghese:


tpolischuk:
https://jira.plos.org/jira/browse/AMBR-514 Testing ticket.

PR #1453, opened 2018-07-18
AMBR-476 Removing the curated FTL include for mobile

tpolischuk:
Corresponding PR in plos-themes:

https://github.com/PLOS/plos-themes/pull/908


tpolischuk:
Updated the PR

PR #1455, opened 2018-07-23
AMBR-501 Using regex to strip the XML from the article title when sha…

tpolischuk:
…ring via e-mail


src/main/java/org/ambraproject/wombat/controller/ArticleController.java
```diff
@@ -731,6 +731,7 @@ public String emailArticle(HttpServletRequest request, HttpServletResponse respo
         .getIngestionMetadata();
 
     String title = articleMetadata.get("title").toString();
+    title = title.replaceAll("<[^>]+>", "");
```
sgatchell: `title` is only used further below in the email message.  This would be a good time to move this code closer to where it is actually used.

PR #1457, opened 2018-08-07
AMBR-556 use https for external sites

sgatchell:
migrate HTTP > HTTPS
- in harcoded links to external resources
- as the default protocol for urls submitted for comments and mediaCuration


anilop:
LGTM!

PR #1464, opened 2018-08-23
Ambr 218 encode math as data uri

egh:
:warning: WORK IN PROGRESS - don't merge

https://jira.plos.org/jira/browse/AMBR-218

An implementation to encode math images as data-uris to reduce the large number (> 100) of requests for math-heavy articles.

The implementation is done as a Saxon function that is used in the stylesheet (see https://github.com/PLOS/wombat/pull/1464/files#diff-43df5e74de352e626062f5ba9acbbf2bR1613)
This function either returns a data-url if it can successfully download and encode the image thumbnail, or it returns a fallback value.


egh:
I am still in the process of testing this locally, but I believe it is ready for review.

PR #1470, opened 2018-09-24
AMBR-649 identify presence of TPR data in an artice

sgatchell:
Read the article's XML to identify the presence of TPR content.  Write a log to the browser console if present.


sgatchell:
obsoleted by https://github.com/PLOS/wombat/pull/1471

PR #1471, opened 2018-09-25
AMBR-631 TPR content in a tab

sgatchell:
If the article has TPR content show a "Peer Review" tab. 

Tab content is bogus.

### DIFF
--- a/src/main/java/org/ambraproject/wombat/controller/ArticleMetadata.java
+++ b/src/main/java/org/ambraproject/wombat/controller/ArticleMetadata.java
@@ -229,7 +229,9 @@ public class ArticleMetadata {
   String getPeerReview() throws IOException {
     return factory.corpusContentApi.readManuscript(articlePointer, site, "html", (InputStream stream) -> {
       String articleXmlStr = IOUtils.toString(stream);
-      return StringUtils.substringBetween(articleXmlStr, "decision-letter");
+
+      String crudeDecisionLetterExtract = StringUtils.substringBetween(articleXmlStr, "decision-letter");
+      return crudeDecisionLetterExtract;
     });
   }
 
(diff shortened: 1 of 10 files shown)


anilop:
Are there any performance implications to reading this data from the XML?  Is it faster to read this from Rhino? Will any other feature or service Query Rhino for a flag to see if n article TRP or not? I'm thinking reporting, tests etc....

src/main/webapp/WEB-INF/themes/desktop/sass/pages/_article.scss
```diff
@@ -411,6 +411,9 @@ $article-sidebar-width: 160;
     a.article-tab-5 {
       background: lighten($tabs-base-color, 40%);
     }
+    a.article-tab-6 {
+      background: lighten($tabs-base-color, 30%);
```
anilop: Shouldn't this be 50%?

sgatchell: I couldn't remember if we were taking the mockup literally with those shades.  It can be 50%

src/main/java/org/ambraproject/wombat/controller/ArticleMetadata.java
```diff
@@ -222,6 +225,16 @@ RevisionMenu getRevisionMenu() throws IOException {
     return new RevisionMenu(revisionList, isDisplayingLatestRevision);
   }
 
+
+  String getPeerReview() throws IOException {
```
sgatchell: Instead of reading XML let's call the rhino API:  http://localhost:8082/articles/10.1371%2B%2Bjournal.pone.0178719/ingestions/1/items

Look for items like this one:
```
"10.1371/journal.pone.0178719.r002": {
      "doi": "10.1371/journal.pone.0178719.r002",
      "itemType": "reviewLetter",
      "files": {
        "letter": {
          "bucketName": "corpus",
          "crepoKey": "10.1371/journal.pone.0178719.r002.xml",
          "crepoUuid": "c391cb75-b3c1-4c7a-8916-70efff5b0e43",
          "size": 23347
        }
      }
    } 
```

sgatchell:
@anilop now using itemTable, pls re-review

sgatchell:
@anilop logging removed, ready for re-review

PR #1473, opened 2018-10-01
Use apt-repo plugin to install dev in /var/tmp/apt

egh:
Only effects development.


sgatchell:
@egh I'm ignorant. why?

egh:
@sgatchell This allows you to run:
```
host os $ mvn package
...
guest os $ sudo apt-get update
...
guest os $ sudo apt-get upgrade
... installs upgraded wombat packages ..
```

Because the `/var/tmp/apt` dir is shared between the host and guest OS and is set up by salt as an "apt repository". Without this change you need to scp & run dpkg -i and jump through other hoops to get a `.deb` package installed in a vagrant.

sgatchell:
Thank you!  Making vagrants better.  :shipit: 

Adding @tpolischuk who may be interested since he's bee n working on the ambra vagrant scene

PR #1488, opened 2019-01-02
V3.7.4rc merge to develop

sgatchell:
We rolled back the release but we still want to merge changes made in the release to develop

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
   <artifactId>wombat</artifactId>
   <name>wombat</name>
   <packaging>war</packaging>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -29,7 +29,7 @@
   <artifactId>wombat</artifactId>
   <name>wombat</name>
   <packaging>war</packaging>
-  <version>3.7.2-SNAPSHOT</version>
+  <version>3.7.4</version>
```
sgatchell: @gt-plos What is the correct value for <version>?  is it going to be `3.7.4-SNAPSHOT`?

PR #1492, opened 2019-01-08
AMBR-810 Fix synchronized statement to avoid a global lock

egh:
*JIRA issue:* https://jira.plos.org/jira/browse/AMBR-810

## What this PR does:

Fixes synchronization to avoid a global lock in some core code.

## Developer Notes

Ad-hoc analysis with jvisualvm showed that java was spending a lot of time in the `getCompiledAssetLink` method. I guessed that this is probably because there is a global lock there.

As background, wombat includes (wrongly, in my opinion) a runtime asset compilation method which compiles the assets and then caches them. Because it would be a bad idea to compile the assets in multiple threads at once, there is a lock around the compilation process. Unfortunately, this lock was also around the code that generates a unique hash to look up whether the compilation needs to occur. Also unfortunately, this was a global lock. This means that this code is essentially single threaded for the entire wombat instance.

It's a bad practice to have a global lock in general, and it's bad practice to lock on more code than is absolutely necessary. I have attempted to fix both these problems: the former by using Guava's Striped implementation https://github.com/google/guava/wiki/StripedExplainedand the latter by moving the lock down around only the build + write of the compiled asset.

The striped lock uses 10 locks and guarantees that when getting a lock for a given key that the same lock will always be returned. I have locked the asset compilation on the key that is used to retrieve the assets, so this should ensure that we never try to compile the same asset at the same time.

# Code Author Tasks:

- [X] Any necessary documentation in confluence or READMEs is updated.

# Code Reviewer Tasks

### DIFF
--- a/src/main/java/org/ambraproject/wombat/service/AssetServiceImpl.java
+++ b/src/main/java/org/ambraproject/wombat/service/AssetServiceImpl.java
@@ -22,9 +22,13 @@
 
 package org.ambraproject.wombat.service;
 
+import java.util.concurrent.locks.Lock;
+
 import com.google.common.base.Preconditions;
+import com.google.common.util.concurrent.Striped;
 import com.yahoo.platform.yui.compressor.CssCompressor;
 import com.yahoo.platform.yui.compressor.JavaScriptCompressor;
+
 import org.ambraproject.rhombat.cache.Cache;
 import org.ambraproject.wombat.config.RuntimeConfiguration;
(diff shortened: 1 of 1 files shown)


mbowen:
@egh  nice job on lock impl and pr notes. this bit of code is tricky enough that almost warrants a test, but suppose we need to draw boundaries to prevent writing too much code (a test for the test for the test ...). again, good job.

PR #1493, opened 2019-01-11
Ambr 632 tpr transform

sgatchell:
*JIRA issue:* https://jira.plos.org/jira/browse/AMBR-632

## What this PR does:
Aggregates peer review XML content and transforms it to HTML which is presented in the Peer Review tab of the article page.

## Developer Notes
The XSL transformation is the complex bit, but the HTML we produce is a simple table which is nice.

# Code Author Tasks:

# Code Reviewer Tasks


src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+      SAXParser sp = SAXParserFactory.newInstance().newSAXParser();
+      xmlReader = sp.getXMLReader();
+    } catch (ParserConfigurationException e) {
+      throw new RuntimeException(e);
+    } catch (SAXException e) {
+      throw new RuntimeException(e);
+    }
+    SAXSource xmlSource = new SAXSource(xmlReader, new InputSource(new StringReader(peerReviewXml)));
+
+    StreamSource xslSource = new StreamSource(new File("src/main/webapp/WEB-INF/themes/desktop/xform/peer-review-transform.xsl"));
```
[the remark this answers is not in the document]
sgatchell: it certainly should

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+    StringWriter htmlWriter = new StringWriter();
+    try {
+      Transformer transformer = TransformerFactory.newInstance().newTransformer(xslSource);
+      transformer.transform(xmlSource, new StreamResult(htmlWriter));
+    } catch (TransformerException e) {
+      throw new RuntimeException(e);
```
sgatchell: We might need something prettier here.  RuntimeException is pretty ugly.

src/main/java/org/ambraproject/wombat/controller/ArticleMetadata.java
```diff
@@ -43,8 +43,7 @@
 import org.springframework.ui.Model;
 
 import javax.servlet.http.HttpServletRequest;
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
```
egh: https://confluence.plos.org/confluence/display/DEV/Java+Coding+Standards no wildcard imports.

sgatchell: Good call.  FYI, you need to add some 9's to intellij to keep it from using wildcard imports: https://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+    SAXSource xmlSource = new SAXSource(xmlReader, new InputSource(new StringReader(allReviewsXml)));
+
+    ClassLoader classLoader = getClass().getClassLoader();
+    InputStream stream = classLoader.getResourceAsStream(xsl);
+    StreamSource xslSource = new StreamSource(stream);
+
+    StringWriter htmlWriter = new StringWriter();
+    try {
```
egh: Is it possible to use https://github.com/PLOS/wombat/blob/develop/src/main/java/org/ambraproject/wombat/service/XmlUtil.java#L114 here?

src/main/java/org/ambraproject/wombat/service/PeerReviewService.java
```diff
@@ -0,0 +1,8 @@
+package org.ambraproject.wombat.service;
+
+import java.io.IOException;
+import java.util.Map;
+
+public interface PeerReviewService {
+  String asHtml(Map<String, ?> itemTable) throws IOException;
```
mbowen: A javadoc comment might be helpful here. I get what return type will be, however, have no idea what an item table is, well a map ... of course, would expect to look at tests for better usage, but should be usable by just inspecting this interface.

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -0,0 +1,175 @@
...
+    String xml = getAllReviewsXml(itemTable);
+    String html = xmlToHtml(xml, DEFAULT_PEER_REVIEW_XSL);
+    return html;
+  }
+
+  /**
+   * Convert peer review XML to HTML
+   *
+   * @param allReviewsXml
+   * @return an HTML representation of peer review content
```
mbowen: should add @param for xsl (xsl template filename)

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -0,0 +1,175 @@
...
+   */
+  String xmlToHtml(String allReviewsXml, String xsl) {
+    if (allReviewsXml == null) return null;
```
mbowen: should probably update guard condition to return null if either xml or xsl is null. should throw an exception instead?

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -0,0 +1,175 @@
...
+
+    XMLReader xmlReader = null;
+    try {
+      SAXParser sp = SAXParserFactory.newInstance().newSAXParser();
+      xmlReader = sp.getXMLReader();
+    } catch (ParserConfigurationException e) {
+      throw new RuntimeException(e);
+    } catch (SAXException e) {
+      throw new RuntimeException(e);
```
mbowen: this could be collapsed into a multi-catch block.
catch (ParserConfigurationException | SAXException e) {
  throw new RuntimeException(e);
}

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
sgatchell: `private static String createXmlString(Node node)` is used privately in XmlUtil.  We'd need to at least make it public and allow it to take a stylesheet as a param.

src/main/resources/peer-review-transform.xsl
```diff
@@ -0,0 +1,105 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xsl:stylesheet version="2.0"
+                xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+>
+  <xsl:output method="html" />
+  <xsl:template match="/">
+    <h2>Peer Review History</h2>
+    <table class="table table-bordered review-history">
+      <tbody>
+        <xsl:for-each select="peer-review/revision">
```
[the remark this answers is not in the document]
sgatchell: I concur.  Let me see what I can do here.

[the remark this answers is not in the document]
mbowen: +1

src/main/resources/peer-review-transform.xsl
```diff
@@ -0,0 +1,105 @@
...
+          <tr>
+            <th>
+              <xsl:if test="position() = 1">
```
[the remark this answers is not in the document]
sgatchell: Less repetition, thank you.

src/test/java/org/ambraproject/wombat/service/PeerReviewServiceTest.java
```diff
@@ -0,0 +1,124 @@
...
+      }
+    };
+
+    String html = serviceWithMockedContent.asHtml(itemTable);
+    assertThat(html, containsString("Original Submission"));
+    assertThat(html, containsString("InitialDecisionLetterSampleBody"));
+    assertThat(html, containsString("Revision 1"));
+//    assertThat(html, containsString("FirstRoundAuthorResponseSampleBody"));
+    assertThat(html, containsString("FirstRoundDecisionLetterSampleBody"));
+
```
mbowen: getAllReviewsXml() in PeerReviewServiceImpl is pretty long. it would be great to have tests which exercise all branches... I'm not sure that current test coverage does that.

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -0,0 +1,175 @@
...
+   * @return an HTML representation of peer review content
+   */
+  String xmlToHtml(String allReviewsXml, String xsl) {
+    if (allReviewsXml == null) return null;
+
+    XMLReader xmlReader = null;
+    try {
+      SAXParser sp = SAXParserFactory.newInstance().newSAXParser();
+      xmlReader = sp.getXMLReader();
+    } catch (ParserConfigurationException e) {
```
egh: This seems to me like you are just working around static exceptions. Is there a way to bubble up or usefully catch these exceptions?

sgatchell: I don't plan to recover from any of these exceptions.

egh: But you are rethrowing them as Runtime exceptions. What will be the effect of that? Maybe I am missing something.

egh: I see. This makes sense to me! Thank you for explaining.

src/main/resources/peer-review-transform.xsl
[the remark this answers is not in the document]
sgatchell: Working on this.  It seems to have exposed a flaw in our XSL.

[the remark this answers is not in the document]
sgatchell: Never mind, there's no flaw in our XSL :stuck_out_tongue:.  

Just noting that the issue with funky tbody and caption elements appears to be that our transformed peer review HTML contains a <caption> element as a descendant of a <table> element and firefox is trying to float it above the table.

src/main/resources/peer-review-transform.xsl
```diff
@@ -0,0 +1,118 @@
...
+          <xsl:otherwise>
+            Revision <xsl:value-of select="position() - 1"/>
+          </xsl:otherwise>
+        </xsl:choose>
+        <div class="date">
+          <span class="decision-date">January 1, 1979</span>
+        </div>
+      </th>
+    </tr>
+    <xsl:if test="sub-article/@article-type = 'author-comment'">
```
egh: You can also use `apply-templates` here:
```
  </tr>
  <xsl:apply-templates/>
</xsl:template>

<xsl:template match="sub-article[@article-type='author-comment']">
  <tr>
    <td>
      <div class="date">
...
</xsl:template>

<xsl:template match="sub-article[@specific-use='decision-letter']">
  <tr>
    <td>
      <div class="decision-letter">
...
</xsl:template>

File: src/main/resources/peer-review-transform.xsl [hunk 14]
```diff
@@ -0,0 +1,118 @@
...
+                Lauren Bianchini, Editor
+              </span>
+            </span>
+            <div class="date">
+              <div class="decision-date">
+                January 1, 1979
+              </div>
+            </div>
+            <div class="decision-letter-body">
+              <xsl:apply-templates select="sub-article[@specific-use = 'decision-letter']/body"/>
```
egh: If you use `apply-templates` per the above comment you won't need to repeat the selector here and can just use `<xsl:apply-templates/>`.

src/main/resources/peer-review-transform.xsl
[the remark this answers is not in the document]
sgatchell: I extracted some xsl:templates. It's more readable now, I think, and idiomatic.

src/main/resources/peer-review-transform.xsl
```diff
@@ -0,0 +1,118 @@
...
+               role="button"
+               aria-expanded="false"
+               aria-controls="decisionLetter">
+              Decision Letter
+            </a>
+            <!-- end trigger for expand and collapse -->
+            -
+            <span class="author">
+              <span class="name">
+                Lauren Bianchini, Editor
```
egh: I assume this is a placeholder for now :)

src/main/resources/peer-review-transform.xsl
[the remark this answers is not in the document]
egh: I actually kind of enjoy doing a little XSLT. It's been a while.

src/main/resources/peer-review-transform.xsl
sgatchell: yep, there's a ticket to populate metadata, and another to make it pretty.  This is more of a "prove it works" ticket.

src/main/resources/peer-review-transform.xsl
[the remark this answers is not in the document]
sgatchell: Yeah, I'm getting into it. Your feedback has been transformative!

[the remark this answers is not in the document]
egh: Aww, thanks! It's a different way of thinking where you kind of go from the bottom up.

src/test/java/org/ambraproject/wombat/service/PeerReviewServiceTest.java
```diff
@@ -0,0 +1,130 @@
...
+            "doi", "10.1371/journal.pone.0207232.t001",
+            "itemType", "table"
+        )
+    );
+    String html = new PeerReviewServiceImpl().asHtml(itemTable);
+    assertNull(html);
+  }
+
+  @Test
+  public void testXslt() {
```
mbowen: How about we delete testXslt() from here. A more robust version will be added in AMBR-657 (testAuthorResponse()).

sgatchell: done!

egh:
@sgatchell @mbowen I don't like being the PR police, but if we have a process we should follow it. This branch was merged without approval in the PR and without all the checkboxes checked.

PR #1498, opened 2019-01-22
AMBR-657: render author response in peer review history.

mbowen:
*JIRA issue:* https://jira.plos.org/jira/browse/AMBR-657

*Related PRs:* AMBR-632

## What this PR does:

Renders author's response in peer review history (which is displayed in new Peer Review tab).

## Developer Notes

# Code Author Tasks:

- [X] There are unit tests sufficient for both positive and negative test cases, test coverage has not decreased as part of this work. 

If I used outside libraries:

# Code Reviewer Tasks

### DIFF
--- a/src/main/resources/peer-review-transform.xsl
+++ b/src/main/resources/peer-review-transform.xsl
@@ -128,8 +128,7 @@
         <div itemprop="reviewBody" class="collapse" id="decisionLetter3">
           <div class="author-response">
             <p>
-              bogus author response text for:
-              <xsl:value-of select="front-stub/article-id"/>
+              <xsl:copy-of select="body"/>
             </p>
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


src/main/resources/peer-review-transform.xsl
```diff
@@ -128,8 +128,7 @@
         <div itemprop="reviewBody" class="collapse" id="decisionLetter3">
           <div class="author-response">
             <p>
-              bogus author response text for:
-              <xsl:value-of select="front-stub/article-id"/>
+              <xsl:copy-of select="body"/>
```
sgatchell: How about this:  
```xml
<xsl:apply-templates select="body"/>
```
Instead of copy-of we're now explicitly handle the elements we expect. See decision letter:
```xml
            <div class="decision-letter-body">
              <p>
                <xsl:apply-templates select="body"/>
              </p>
            </div>
```

mbowen: @sgatchell I've updated stylesheet to use apply-templates. updated test too (which now makes fewer assertions).

PR #1502, opened 2019-01-25
AMBR-822 explicitly use saxon transformer

sgatchell:
*JIRA issue:* https://jira.plos.org/jira/browse/-AMBR-822

## What this PR does:
Explicity uses an XML transformer that supports XSL transforms.  We were using a default transformer before this change, but we found that on tomcat 8 with the cargo runner the default transformer does not seem to support XSL.  As a result the Peer Review content is a wall unformatted text.

## Developer Notes
This issue is not present in the [dev environment](https://journals-dev.plos.org/plosone/article/peerReview?id=10.1371/journal.pone.0207232)
Noting that we still don't know what exactly would need to be changed in the local cargo-run environment to have parity with dev in regards to the default transformer. We have chosen not to prioritize further investigation.

# Code Author Tasks:

# Code Reviewer Tasks

### DIFF
--- a/src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
+++ b/src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
@@ -75,7 +75,7 @@ public class PeerReviewServiceImpl implements PeerReviewService {
 
     StringWriter htmlWriter = new StringWriter();
     try {
-      Transformer transformer = TransformerFactory.newInstance().newTransformer(xslSource);
+      Transformer transformer = SiteTransformerFactory.newTransformerFactory().newTransformer(xslSource);
       transformer.transform(xmlSource, new StreamResult(htmlWriter));
     } catch (TransformerException e) {
       throw new RuntimeException(e);
--- a/src/main/java/org/ambraproject/wombat/service/SiteTransformerFactory.java
+++ b/src/main/java/org/ambraproject/wombat/service/SiteTransformerFactory.java
@@ -43,7 +43,7 @@ import java.util.concurrent.ConcurrentHashMap;
 
 public class SiteTransformerFactory {
 
-  private static TransformerFactory newTransformerFactory() {
+  static TransformerFactory newTransformerFactory() {
     // This implementation is required for XSLT features, so just hard-code it here
     // Preferred over TransformerFactory.newInstance because Java system properties can burn in hell
     return new net.sf.saxon.TransformerFactoryImpl();


sgatchell:
Too tedious.  Instead of guessing what a person meant when they checked a box or not I've updated my note:

#### All tasks have been considered.  A box is checked if I took action. If a box is unchecked I consider it irrelevant.

sgatchell:
So, I thought about this a little while getting lunch.  I would prefer a situation where it is assumed that all tasks have been considered, and a checked box indicates action has been taken. Unchecked indicates only the absence of action, and by not deleting it that intent is preserved and the reviewer can ask about the motivation for not taking action, such as irrelevant, too expensive, etc.

sgatchell:
Done as in considered or done as in action was taken?

You don't have to ask either way, but by deleting the checkboxes you don't have a record of what tasks were rejected.

I'll delete the tasks that I took no action on, although I think we could avoid this if we had a shared agreement that an unchecked checkbox means that a task was considered but did not result in any action. I think what's holding us back is that the unchecked box is currently assumed to mean the task was not considered.

egh:
@sgatchell Thanks, Shawn. I was just about to comment that I think we need to come to some shared agreement here. I already have an agenda item for next week's tech chat to discuss this.

sgatchell:
Cool If we could assume that all tasks are considered then we'd have one less state to try to squeeze out of a checkbox.

egh:
Also,

> Done as in considered or done as in action was taken?

Exactly! My thinking is, if the checklist item is deleted/struck through, you considered it but it didn't need to be done. If the checklist item is checked, I know you did it. If it's unchecked, I assume you know it needs to be done, but you haven't done it yet. We need a shared understanding here :)

sgatchell:
I'd prefer an environment that avoids strikethrough/deletion to convey a third state, where instead it is assumed all tasks are considered.
Unchecked: considered and action not needed
Checked: considered and action taken

sgatchell:
...and regular attempts to

egh:
> Unchecked: considered and action not needed

This does not comport with my experience of what happens. IME the box is left unchecked because it was *not* considered.

In any case, thank you for deleting the checkboxes!

sgatchell:
...not close the PR!  and try to winnow down the checkboxes

sgatchell:
> IME the box is left unchecked because it was not considered.

That's the crux of it. Simplifying the PR form would be an ancillary benefit of establishing that tasks are always considered (as part of our code of conduct?).

PR #1505, opened 2019-01-29
Ambr 634 tpr mobile

samV-plos:
*JIRA issue:* https://jira.plos.org/jira/browse/AMBR-842

## What this PR does:

This PR provides the HTML and CSS prototype for the new Peer Review screen for _mobile_. 
This does not include the accordion (expand/collapse) functionality.

*Please note*
I added a temporary CSS rule for the collapsed/expanded content, so that it would be easier to distinguish what is supposed to be collapsed on initial view.  The text currently appears in grey.

-------------------
plos one dosimetric and radiobiological comparison in different dose calculation grid sizes between acuros xb and anisotropic analytical algorithm for prostate vmat

# Code Author Tasks:

- [ ] There are unit tests sufficient for both positive and negative test cases, test coverage has not decreased as part of this work. 

# Code Reviewer Tasks


egh:
@samV-plos These changes are breaking some tests. I'm happy to fix the tests for you if you like.

sgatchell:
@egh I was just looking into the broken tests.  I'm planning to merge develop and fix them.

src/main/webapp/WEB-INF/themes/mobile/resource/css/content.css
```diff
@@ -3,24 +3,33 @@
...
+/*#content p, 
+#content ul, 
+#content ol, 
+#content table{
+  font-size: 14px;
   line-height: 18px;
   margin-bottom: .5em;
   margin-top: .5em;
-  }
+  }*/
```
sgatchell: Are you saving these for later?

src/main/webapp/WEB-INF/themes/mobile/resource/css/content.css
```diff
@@ -57,7 +66,7 @@ font-size: 2.25rem;
 
 #content table tr {
   border-bottom: 1px solid #999;
-  }
+  }*/
```
sgatchell: should it stay or should it go?

src/main/webapp/WEB-INF/themes/mobile/resource/css/peer-review.css
```diff
@@ -0,0 +1,110 @@
...
+  width: 100%;
+  display: block;
+}
+
+.decision__date {
+  font-weight: 400;
+  text-indent: 1rem;
+}
+
+.decision-letter__title {
```
sgatchell: @samV-plos I expect there to exist analogous `author-comment...` and `acceptance-letter...` classes. How do you want to handle the css classes so that they are all styled the same as `decision-letter...`?

src/main/webapp/WEB-INF/themes/mobile/resource/css/content.css
samV-plos: No, I'll fix ;-)

src/main/webapp/WEB-INF/themes/mobile/resource/css/peer-review.css
samV-plos: yes, we should normalize these. 
I realized that some of the content was keying off the classes in the mark-up, so wasn't sure if they were ok to change.

samV-plos:
@sgatchell @egh --

- Can someone show me how can I check these failing tests myself? Locally, everything seems to compile. 
- It seems like we're coupling the data layer and the presentation layer too closely, and I think it would be good to come up with a class naming convention that helps us identify which classes are purely presentational vs. which ones have a dependency, so that anyone working on the front-end will be able to discern between the two.

For instance, I noticed that the code is looking for a "review-history" class, as well as "decision-letter" and "author-response".  I'm assuming that a change to those classes will cause some tests to fail.  Without comments in the mark-up, or some kind of naming convention rules, there's no way to protect these templates from similar failings in the future.

We should figure out a prefix for these class names, that can be 1) documented and 2) easily identified by any developer (front-end or back-end) that is new to the project.  For example:
- xsl-review-history or data-review-history or java-review-history (whatever most accurately represents the class function, as long as it is differentiated from the presentation layer)
- xsl- author-response
- xsl- decision-letter

Does this seem reasonable?  Comments or objections?

sgatchell:
@samV-plos Generalizing the classes from `decision-letter-X` and `author-response-X` to `.letter--X` is appropriate where they share styles.  Fixing the tests to match the reality, as you did, is IMO the simplest approach and reasonable, both now and in the future.

I'm adding the acceptance letter in AMBR-833 and I expect these generalizations to make it more likely to "just work".

samV-plos:
@sgatchell I'll handle the footer issue in a separate ticket:
https://jira.plos.org/jira/browse/AMBR-837

PR #1509, opened 2019-02-07
Ambr 831 accordion

sgatchell:
*JIRA issue:* https://jira.plos.org/jira/browse/AMBR-831

## What this PR does:

Individual letters are expandable and start out collapsed, both in mobile and desktop.
 
## Developer Notes
The mobile theme implements a homegrown accordion used to present article sections. I elected to create a separate accordion script for peer review rather than try to refactor the existing homegrown accordion because it was tightly bound to the implementation of `MainAccordion` as a list.
see: https://github.com/PLOS/wombat/blob/develop/src/main/webapp/WEB-INF/themes/mobile/resource/js/navigation.js#L208
 
# Code Reviewer Tasks


samV-plos:
Looks beautiful & works great!  I'm good with moving the desktop SCSS into its own file.
There are a few minor things left to do (change the body text back to black, remove the [ brackets ] ), but I can tackle those in a separate ticket.

src/main/resources/peer-review-transform.xsl
```diff
@@ -103,41 +99,35 @@
...
+    <tr class="peer-review-accordion-item">
       <td class="letter">
         <div class="author-response">
-          <!-- author response -->
           <time class="letter__date">
             <xsl:value-of select=".//named-content[@content-type = 'author-response-date']" />
           </time>
           <div class="letter__title">
-            <a class="letter--toggle" data-toggle="collapse" href="#decisionLetter2">[Author Response]</a>
+            <a class="peer-review-accordion-expander" href="#">[Author Response]</a>
```
samV-plos: Can you change 
`[Author Response]` to `Author Response`
(and remove from the other toggles)

sgatchell: will do, thanks!

PR #1513, opened 2019-02-12
AMBR-835: TPR Metadata

mbowen:
*JIRA issue:* https://jira.plos.org/jira/browse/AMBR-835

*Related PRs:* 

## What this PR does:

This PR fetches and displays the specified metadata in the peer review history.
Specifically, it gets the "received date" (aka, original submission date) from
the article xml. The editor and author/decision letter dates are retrieved
from the TPR payload.

## Developer Notes

I suspect this PR will need at least one more round of edits before accepting
and merging. Here are some ponderables. Reviewer may discover other issues.

1. Should we rename the received-date tag? article-received-date seems a better
name, or other ...

2. I put received date inside of the peer-review xml block, but maybe it should go
elsewhere. you'll see some explicit references to this node in the template. i
defined a no-op template matcher to ignore this node which may be a smell to do something else.

3. dates are rendered in "dd MMM yyyy" format (9 Oct 2018), however, codepen
mock-up specifies a more verbose format (October 9, 2018). I hesitated to change
this because date is in the shorter format in the TPR payload. Whatever we do, seems
we want to be consistent here.

4. I did most of my testing locally via unit tests, however, did package up and
deploy to my vagrant running Tomcat 7. In preparation for Ubuntu upgrade, Wombat
Tomcat has been upgraded to 8.5; this has been merged into develop. This shouldn't be
a problem, I just haven't tested this nor merged these changes into this branch.

# Code Author Tasks:

If I used outside libraries:

# Code Reviewer Tasks

### DIFF
--- a/src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
+++ b/src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
@@ -2,49 +2,43 @@ package org.ambraproject.wombat.service;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.MoreCollectors;
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.StringReader;
-import java.io.StringWriter;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.TreeMap;
-import javax.xml.parsers.ParserConfigurationException;
+import org.ambraproject.wombat.service.remote.ContentKey;
+import org.ambraproject.wombat.service.remote.CorpusContentApi;
+import org.apache.commons.lang.CharEncoding;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.util.EntityUtils;
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Document;
+import org.jsoup.nodes.Element;
+import org.jsoup.nodes.Entities;
+import org.jsoup.parser.Parser;
+import org.jsoup.select.Elements;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
+import org.xml.sax.XMLReader;
 
-import java.time.LocalDate;
-import java.time.format.DateTimeFormatter;
-import java.time.format.DateTimeParseException;
-import java.time.format.FormatStyle;
-import java.util.UUID;
+import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.parsers.SAXParser;
 import javax.xml.parsers.SAXParserFactory;
 import javax.xml.transform.Transformer;
 import javax.xml.transform.TransformerException;
-import javax.xml.transform.TransformerFactory;
 import javax.xml.transform.sax.SAXSource;
 import javax.xml.transform.stream.StreamResult;
 import javax.xml.transform.stream.StreamSource;
-import org.xml.sax.XMLReader;
-
-import org.apache.http.client.methods.CloseableHttpResponse;
-
-import org.apache.http.util.EntityUtils;
-
-import org.jsoup.Jsoup;
-import org.jsoup.nodes.Document;
-import org.jsoup.nodes.Element;
-import org.jsoup.parser.Parser;
-import org.jsoup.select.Elements;
-
-import org.springframework.beans.factory.annotation.Autowired;
-
-import org.ambraproject.wombat.service.remote.ContentKey;
-import org.ambraproject.wombat.service.remote.CorpusContentApi;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.StringReader;
+import java.io.StringWriter;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.time.format.FormatStyle;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.UUID;
 
 
 public class PeerReviewServiceImpl implements PeerReviewService {
@@ -124,6 +118,8 @@ public class PeerReviewServiceImpl implements PeerReviewService {
       xml = xml.replaceAll("<\\?xml(.+?)\\?>", "");
 
       Document parsedXml = Jsoup.parse(xml, "", Parser.xmlParser());
+      parsedXml.outputSettings().escapeMode(Entities.EscapeMode.xhtml);
+      parsedXml.outputSettings().charset(CharEncoding.UTF_8);
 
       // re-format review dates (9 Oct 2018 -> October 9, 2018)
       Elements reviewDates = parsedXml.select("named-content[content-type~=(letter-date|author-response-date)]");
@@ -167,7 +163,7 @@ public class PeerReviewServiceImpl implements PeerReviewService {
     return peerReviewContent;
   }
 
-   /**
+  /**
    * Get received date (aka, original submission date) from article manuscript.
    * @param itemTable
    * @return article received date (as a string)
--- a/src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
+++ b/src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
@@ -18,6 +18,7 @@ import org.xml.sax.SAXException;
 
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
 import java.time.format.FormatStyle;
 import java.util.UUID;
 import javax.xml.parsers.SAXParser;
@@ -48,6 +49,11 @@ import org.ambraproject.wombat.service.remote.CorpusContentApi;
 public class PeerReviewServiceImpl implements PeerReviewService {
 
   public static final String DEFAULT_PEER_REVIEW_XSL = "peer-review-transform.xsl";
+
+  private static DateTimeFormatter DATE_FORMATTER = DateTimeFormatter.ofPattern(""
+      + "[M d yyyy]"      // 9 15 2010
+      + "[d MMM yyyy]"    // 15 Sep 2010 
+    );
 
   @Autowired
   private CorpusContentApi corpusContentApi;
@@ -183,11 +189,25 @@ public class PeerReviewServiceImpl implements PeerReviewService {
       XmlUtil.extractText(XmlUtil.extractElement(receiveDateXml, "day")),
       XmlUtil.extractText(XmlUtil.extractElement(receiveDateXml, "year")));
 
-    DateTimeFormatter parseFormatter = DateTimeFormatter.ofPattern("M d yyyy");
-    LocalDate receivedDate = LocalDate.parse(receivedDateString, parseFormatter);
+    return formatDate(receivedDateString);
+  }
 
-    // ex: January 1, 2018
-    return DateTimeFormatter.ofLocalizedDate(FormatStyle.LONG).format(receivedDate);
+  /**
+   * Formats recognized date string into a full month name format 
+   * (ex: January 1, 2010). It will strip all needless whitespace before
+   * processing.
+   *
+   * @param dateString date string to format.
+   * @return formatted date if recognized, else the empty string.
+   */
+  String formatDate(String dateString) {
+    try {
+      String normalizedDate = dateString.replaceAll("\\s{2,}"," ").trim();
+      return DateTimeFormatter.ofLocalizedDate(FormatStyle.LONG)
+        .format(LocalDate.parse(normalizedDate, DATE_FORMATTER));
+    } catch (DateTimeParseException e) {
+      return "";
+    }
   }
 
   /**
--- a/src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
+++ b/src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
@@ -37,6 +37,7 @@ import org.apache.http.util.EntityUtils;
 
 import org.jsoup.Jsoup;
 import org.jsoup.nodes.Document;
+import org.jsoup.nodes.Element;
 import org.jsoup.parser.Parser;
 import org.jsoup.select.Elements;
 
@@ -123,6 +124,13 @@ public class PeerReviewServiceImpl implements PeerReviewService {
       xml = xml.replaceAll("<\\?xml(.+?)\\?>", "");
 
       Document parsed = Jsoup.parse(xml, "", Parser.xmlParser());
+
+      // re-format letter date (9 Oct 2018 -> October 9, 2018)
+      Element letterDate = parsed.select("named-content[content-type$=letter-date]").first();
+      if(letterDate != null) {
+        letterDate.text(formatDate(letterDate.text()));
+      }
+
       Elements subArticle = parsed.select("sub-article");
       if(!subArticle.isEmpty() && subArticle.attr("specific-use").equals("acceptance-letter")){
         acceptanceLetter = xml;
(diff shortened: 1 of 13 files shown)


src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -134,11 +144,34 @@ String getAllReviewsAsXml(List<Map<String, ?>> reviewLetterItems) throws IOExcep
...
   }
 
+   /**
+   * Get article metadata to include in peer review history, such as the
+   * received date (aka, the original submission date).
+   *
+   * @param itemTable
+   * @return map with article metadata. 
+   */
+  Map<String,String> getArticleMetadata(Map<String,?> itemTable) throws IOException {
```
sgatchell: This name could confuse, given that there already exists an `ArticleMetadata` class that gets metadata from the rhino DB. Maybe call it `getArticleXmlData` or just get the date since that's all you really need.

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -134,11 +144,34 @@ String getAllReviewsAsXml(List<Map<String, ?>> reviewLetterItems) throws IOExcep
...
+    Map<String,String> metadata = new HashMap<>();
+
+    for (Object itemObj : new TreeMap(itemTable).values()) {
+
+      if (((Map<String,?>) itemObj).get("itemType").equals("article")) {
+        String articleXml = getContentXml((Map<String,?>)itemObj, "manuscript");
+        metadata.put("receivedDate", getReceivedDate(articleXml));
+        break;  //TODO: at most 1 article type in payload?
```
sgatchell: I believe it's safe to assume 1 article

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -134,11 +144,34 @@ String getAllReviewsAsXml(List<Map<String, ?>> reviewLetterItems) throws IOExcep
...
+   * @param itemTable
+   * @return map with article metadata. 
+   */
+  Map<String,String> getArticleMetadata(Map<String,?> itemTable) throws IOException {
+    Map<String,String> metadata = new HashMap<>();
+
+    for (Object itemObj : new TreeMap(itemTable).values()) {
+
+      if (((Map<String,?>) itemObj).get("itemType").equals("article")) {
+        String articleXml = getContentXml((Map<String,?>)itemObj, "manuscript");
```
sgatchell: https://github.com/PLOS/wombat/blob/develop/src/main/java/org/ambraproject/wombat/service/ArticleServiceImpl.java#L77

Here's an example where we get the article.  You might want to do something like this.  It implies the expectation of only one article.  Then you could get the date, and fail early if the article was unexpectedly not found.

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -174,7 +207,29 @@ String getReviewXml(Map<String, ?> metadata) throws IOException {
...
+
+    String receiveDateXml = XmlUtil.extractElement(xml, "date", "date-type", "received");
+
+    String receivedDateString = String.format("%s %s %s",
+      XmlUtil.extractText(XmlUtil.extractElement(receiveDateXml, "month")),
+      XmlUtil.extractText(XmlUtil.extractElement(receiveDateXml, "day")),
+      XmlUtil.extractText(XmlUtil.extractElement(receiveDateXml, "year")));
+
+    DateTimeFormatter parseFormatter   = DateTimeFormatter.ofPattern("M d yyyy");
+    DateTimeFormatter displayFormatter = DateTimeFormatter.ofPattern("dd MMM yyyy");
```
sgatchell: You might want to move this closer to where it's used.

src/main/resources/peer-review-transform.xsl
```diff
@@ -33,29 +33,32 @@
       </p>
     </div>
   </xsl:template>
-  
+
   <xsl:template match="revision">
+    <xsl:variable name="revision-row">
```
sgatchell: I gotta admit I have no idea how this works as a substitute revision counter.  I'll have to look it up.

src/main/java/org/ambraproject/wombat/service/XmlUtil.java
```diff
@@ -90,6 +98,11 @@ public static String extractElement(String xml, String tagName) throws IOExcepti
     return createXmlString(extractElement(createXmlDocument(xmlSource), tagName));
   }
 
+  public static String extractElement(String xml, String tagName, String attrName, String attrValue) throws IOException {
```
sgatchell: nice utility!

src/main/resources/peer-review-transform.xsl
sgatchell: looked up `xsl:number`, and now I know.  :)

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -134,11 +144,34 @@ String getAllReviewsAsXml(List<Map<String, ?>> reviewLetterItems) throws IOExcep
     // append the acceptance letter
     peerReviewContent += acceptanceLetter;
 
+    peerReviewContent += "<received-date>" + articleMetadata.get("receivedDate") + "</received-date>";
```
sgatchell: Re: dev note 1.

I would name the tag as close as we can to the corresponding tag in the article XML, for easier troubleshooting.  `article-received-date` is more precise than `received-date` but I don't feel strongly about it.  Your call.

src/main/resources/peer-review-transform.xsl
```diff
@@ -189,4 +200,9 @@
   </xsl:template>
   
   <xsl:template match="named-content" />
+
+  <!-- ignore received date found during template processing.
+       date is referenced directly in revision #0 block --> 
+  <xsl:template match="received-date" />
```
sgatchell: re: dev note 2:
I think this sort of no-op is a common pattern where you want to present the value of something in a different location but you don't want the value to appear again in it's original position due to it passing through via `<apply-templates>`.  My XSL experience is limited but I don't consider it a smell.

These no-ops might even make up the majority of the article transform XSL.

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
sgatchell: re: dev note 3: 
This is the only place where the date is received in parts.  We're pulling all other dates as literal strings from the review XML.  I would not be surprised if Tina has specified the date format in the instructions for Apex, but it would be a good idea to ask @ramzizaz and Katie, and also ask @samV-plos if she intended to specify a particular format or wants to conform with what we receive from Apex.

sgatchell:
re: dev note 4:
Well, I merged develop so that's behind us now. I've noticed that running `mvnDebug` with the cargo runner doesn't seem to work for me.  I didn't find a quick fix and switched to print statements instead.

We're going to have a bunch of end-to-end testing for the TPR pipeline, see AMBR-625.  Having just a few unit tests that aren't too fragile seems to me least costly to maintain.

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
samV-plos: I'm using microdata in the markup, to enhance machine-readability.  The value of the datetime attribute is the date specified using YYYY-MM-DD format:
`<time datetime="2011-04-01">04/01/11</time>`

sgatchell: I've only been populating the nodes that humans will see.  To that end, @samV-plos  your prototype indicates a format of  `MMMM d, yyyy` - e.g. `October 4, 2018` 

I've asked Tina in the ticket to confirm that format we see in her sample XML is what we should expect in the sub-article XML:  `d MMM yyyy` - e.g. `4 Oct 2018`.

@samV-plos we should create a new ticket to populate the nodes that are not intended for visual consumption. I am certain that those nodes have not been implemented to your expectations.

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
mbowen: I've simplified to just fetching received date which may be all we ever need from article (YAGNI, as you said).

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
mbowen: Ok.

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
mbowen: Beautiful. I wanted to do something like this instead of brute-force iteration. have replaced with this.

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
mbowen: Moved.

src/main/resources/peer-review-transform.xsl
mbowen: Actually, the position() code looked right, but I never saw the header row (row 1). when put in some debug values, I saw the revision #0 row had position()=2,  and revision #1 had position 4 - wtf. I read up on provision which documents that it numbers elements in a node set. I then experimented with xsl:number and got expected results. I think we could use position() but would need to modify match node set.

src/main/java/org/ambraproject/wombat/service/XmlUtil.java
mbowen: was surpised this hadn't already been implemented, but suspect this utility implements just enough. so, added a bit more.

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
mbowen: renamed to article-received-date.

src/main/resources/peer-review-transform.xsl
mbowen: ah. ok. I really don't have that much experience with xsl. let me rephase. smell is actually a pleasant aroma.

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -144,32 +147,49 @@ String getAllReviewsAsXml(List<Map<String, ?>> reviewLetterItems, Map<String,?>
...
+
+    String receiveDateXml = XmlUtil.extractElement(xml, "date", "date-type", "received");
+
+    String receivedDateString = String.format("%s %s %s",
+      XmlUtil.extractText(XmlUtil.extractElement(receiveDateXml, "month")),
+      XmlUtil.extractText(XmlUtil.extractElement(receiveDateXml, "day")),
+      XmlUtil.extractText(XmlUtil.extractElement(receiveDateXml, "year")));
+
+    DateTimeFormatter parseFormatter   = DateTimeFormatter.ofPattern("M d yyyy");
+    DateTimeFormatter displayFormatter = DateTimeFormatter.ofPattern(APEX_DATE_FORMAT);
```
sgatchell: The display format is Sam's spec, not Apex's.  You haven't yet had to deal with the Apex date format since the article received date is formatted in parts.  

It's the dates for the letters and author responses that come in as strings.  I had populated them for funsies but now we'll need to make sure they get parsed and formatted correctly.

This date formatting, and handling a letter that does not have exactly one editor, are requirements that I wasn't considering in the previous review.

mbowen:
@sgatchell I have made most of the changes. there is still one issue, maybe two that still need to address.

issue #1. error handling if item type not found. we haven't handled this at all, and maybe don't need too. that is, would crash but should never happen and if does crashing is appropriate. that is, human needs to get involved, and likely need to fix data.

issue #2. we are rendering article received date in "9 Oct 2018" format... it seems we want this format to be consistent with other dates which are display "as is" (ie, from content in xml). it seems this is the APEX format, but maybe not always?

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -144,32 +147,49 @@ String getAllReviewsAsXml(List<Map<String, ?>> reviewLetterItems, Map<String,?>
...
+      .collect(MoreCollectors.onlyElement());
 
-      if (((Map<String,?>) itemObj).get("itemType").equals("article")) {
-        String articleXml = getContentXml((Map<String,?>)itemObj, "manuscript");
-        metadata.put("receivedDate", getReceivedDate(articleXml));
-        break;  //TODO: at most 1 article type in payload? 
-      }
-    }
-    return metadata;
+    String articleXml = getContentXml(articleItem, "manuscript");
```
sgatchell: This method feels much more expressive to me.  Nice!

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
sgatchell: probably a good idea to move this formatter closer to where it is used to avoid confusion.

src/main/resources/peer-review-transform.xsl
sgatchell: Thanks for the fix!  So, the revision number we displayed must have been coincidentally working for one round of revision?

mbowen: i never saw the header row. that is, first revision tag was at position 2.

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -134,11 +156,64 @@ String getAllReviewsAsXml(List<Map<String, ?>> reviewLetterItems) throws IOExcep
...
+
+    return getReceivedDate(articleXml);
+  }
+
+  /**
+   * Extract received date from article xml.
+   * @param xml article xml
+   * @return received date.
+   */
+  String getReceivedDate(String xml) throws IOException {
```
sgatchell: Given that this is referenced once from a method with a similar name, I recommend inlining this method.

src/main/resources/peer-review-transform.xsl
```diff
@@ -85,6 +90,10 @@
     </tr>
   </xsl:template>
 
+  <xsl:template match="contrib">
+    <xsl:value-of select="concat(normalize-space(./name/given-names),' ',normalize-space(./name/surname),', Editor')" />
```
sgatchell: :100:

src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java
```diff
@@ -92,26 +104,36 @@ String transformXmlToHtml(String allReviewsXml, String xsl) {
...
-      String xml = getReviewXml(reviewLetterItem);
+      String xml = getContentXml(reviewLetterItem, "letter");
 
       // strip the XML declaration, which is not allowed when these are aggregated
       xml = xml.replaceAll("<\\?xml(.+?)\\?>", "");
 
-      Document parsed = Jsoup.parse(xml, "", Parser.xmlParser());
-      Elements subArticle = parsed.select("sub-article");
+      Document parsedXml = Jsoup.parse(xml, "", Parser.xmlParser());
+      parsedXml.outputSettings().escapeMode(Entities.EscapeMode.xhtml);
```
sgatchell: Looks like these are chainable.  Let's set the syntax setting too.
```
      parsedXml.outputSettings().escapeMode(Entities.EscapeMode.xhtml).
          syntax(Document.OutputSettings.Syntax.xml).
          charset(CharEncoding.UTF_8);
```

sgatchell: Feels odd using an html parsing library that also does xml, but the api is so simple!

sgatchell: @mbowen check it, proof of concept.  I'm going to refine this and add it to the PR: https://github.com/PLOS/wombat/commit/3d6dbb0e6c412185eb7a7ebc5c731cd54861fb1e

PR #1520, opened 2019-02-22
AMBR-886 Fix issue with upgrade to tomcat8

egh:
In tomcat8, presumably because it supports a different version of the servlet spec, the type of the variables passed in to freemarker can be either a String[] or an AbstractList<String>.

*JIRA issue:* https://jira.plos.org/jira/browse/AMBR-886

## What this PR does:

Handles both cases.

## Testing

To test, run:

`mvn cargo:run`

locally and visit:

http://localhost:8080/plosone/browse/geographical_locations?resultView=cover

If this works, you should not see an error.

You may need to configure your wombat properly. Check with me if your current config does not work.

# Code Author Tasks:

# Code Reviewer Tasks

### DIFF
--- a/src/main/java/org/ambraproject/wombat/freemarker/ReplaceParametersDirective.java
+++ b/src/main/java/org/ambraproject/wombat/freemarker/ReplaceParametersDirective.java
@@ -39,9 +39,10 @@ import org.apache.http.client.utils.URLEncodedUtils;
 import org.apache.http.message.BasicNameValuePair;
 
 import java.io.IOException;
+import java.util.AbstractList;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/ambraproject/wombat/freemarker/ReplaceParametersDirective.java
```diff
@@ -96,15 +97,8 @@ public void execute(Environment environment, Map params, TemplateModel[] loopVar
 
     // The map is passed in as a Map<String, String[]>, but Freemarker doesn't support generics
```
mbowen: Update comment.

egh: I think this is still an accurate comment, see https://github.com/apache/tomcat85/blob/b21d02042c408053a1944b27c3196ec9be3db381/java/org/apache/catalina/connector/RequestFacade.java#L432

egh: I tried for a few hours to figure out why this is happening, but I can't figure it. There is some change between tomcat versions that causes freemarker to pass in a different type as the parameterMap. Frankly all this code is not great: the `replaceParameters` directive is a hack, and we are implementing our own access to the the request parameters to work around a "bug" that probably doesn't effect us (https://github.com/plos/wombat/blob/AMBR-886/src/main/java/org/ambraproject/wombat/controller/SearchController.java#L326).

egh:
@mbowen I updated this fix to support both a `String[]` or an `AbstractList<String>`

mbowen:
nice tests.

PR #1530, opened 2019-03-16
AMBR-905 Fix jackson dependencies for new spring version

egh:
*JIRA issue:* https://jira.plos.org/jira/browse/AMBR-905

## What this PR does:

Upgrades jackson to a version compatible with our spring version.

The underlying stack trace here was being hidden by an error in our error handler. I was able to get to the original stack trace by commenting out the exception handler.

The underlying error was:
```
org.springframework.web.util.NestedServletException: Handler dispatch failed; nested exception is java.lang.NoSuchMethodError: com.fasterxml.jackson.databind.ObjectWriter.forType(Lcom/fasterxml/jackson/databind/JavaType;)Lcom/fasterxml/jackson/databind/ObjectWriter;
```

## Developer Notes

The core of this fix is in 308eeeeb8.

The rest is test cleanup, and adding a regression test.

## To test:

Checkout a969d5f0e and run `mvn clean test`. You should get a test failure - this is expected, because the test here is to ensure that we do not regress this problem with the taxonomy browser.

Now checkout 308eeeeb8, and run `mvn clean test`. The tests should pass.

## Further test notes

I believe the file https://github.com/PLOS/wombat/blob/308eeeeb85c763765bee15b72f22aeef49bab46e/src/test/java/org/ambraproject/wombat/controller/TaxonomyControllerTest.java can provide an exemplar for tests going forward.

Note the configuration inner class: https://github.com/PLOS/wombat/blob/308eeeeb85c763765bee15b72f22aeef49bab46e/src/test/java/org/ambraproject/wombat/controller/TaxonomyControllerTest.java#L52 which sets up a single, non-mocked controller. The rest of the mocked beans are provided by the `ControllerTest` superclass. (In the future we could abstract out more of this into a more generic class.) The required instances are set up by Spring with `@Autowired`. I hope it makes sense. I reworked most of the other controller tests as well using this pattern.

# Code Author Tasks:

__
# Code Reviewer Tasks


src/test/java/org/ambraproject/wombat/controller/StaticResourceControllerTest.java
```diff
@@ -157,17 +173,119 @@ public String describeSource() {
 
   @Configuration
   @EnableWebMvc
-  @Import(WombatControllerTestConfig.class)
```
sgatchell: I got the impression that the WombatControllerTestConfig contained a bunch of generally applicable test setup, but if it was only being used in one place I guess it has limited applicability and it makes sense to inline it.

src/test/java/org/ambraproject/wombat/controller/ControllerTest.java
```diff
@@ -22,183 +22,42 @@
...
-  @Bean
-  protected SiteSet siteSet(Site site) {
-    final SiteSet siteSet = mock(SiteSet.class);
-    when(siteSet.getSites()).thenReturn(ImmutableSet.of(site));
-    return siteSet;
-  }
-
   @BeforeMethod(alwaysRun = true)
-  public void setUp() throws IOException {
+  public void setupMockMvc() throws IOException {
```
sgatchell: more accuracy, thx

src/test/java/org/ambraproject/wombat/controller/PeerReviewControllerTest.java
```diff
@@ -54,17 +53,12 @@ private void setup() throws IOException {
     doReturn(articleMetadata).when(articleMetadataFactory).newInstance(any(), any(), any(), any(), any(), any());
   }
     
-  @Test
+  @Test(expectedExceptions = NotFoundException.class)
```
sgatchell: :+1: I don't think I was aware of this annotation at the time.

src/test/java/org/ambraproject/wombat/controller/ControllerTest.java
sgatchell: oh, and avoids collision

src/test/java/org/ambraproject/wombat/controller/StaticResourceControllerTest.java
egh: Yeah, I figure it's better to put that in the `ControllerTestConfig` that is automatically imported when you extend `ControllerTest`.

src/test/java/org/ambraproject/wombat/controller/ArticleControllerTest.java
```diff
@@ -22,114 +22,81 @@
...
-  @Bean
-  protected CorpusContentApi corpusContentApi() {
-    final CorpusContentApi corpusContentApi = mock(CorpusContentApi.class);
-    return corpusContentApi;
-  }
+  @Autowired
+  ArticleService articleService;
 
-  @Bean
-  protected PeerReviewService peerReviewService() {
```
sgatchell: When I pulled apart ArticleController I was annoyed that this bean definition was still needed here.  Looks like you fixed that by making a spy of ArticleMetadata.Factory.  thx!

egh:
@sgatchell Yes, that should work as long as the class is configured as a mock by the `ControllerTestConfiguration`. The model I settled on was using that class to `mock` almost every required bean, and overriding the configuration to wire up the class under test, e.g at https://github.com/PLOS/wombat/blob/develop/src/test/java/org/ambraproject/wombat/controller/ArticleControllerTest.java#L63

It was the best thing I could come up with to implement the pattern that almost everything should be mocked out except the class being tested and using the least amount of boilerplate code in the test class.

egh:
Oh, also, I wanted to split up the configuration classes (those with `@Bean` annotations) from the test classes with `@Autowired` annotations. Although they can be combined I found the pattern confusing.

sgatchell:
I was certainly confused and intimidated.  I expect future tests to be much quicker to develop.

egh:
I hope so, it sure was a mess. It's a little cleaner now, I think,.

PR #1531, opened 2019-03-22
AMBR-857 show review doi link

sgatchell:
*JIRA issue:* https://jira.plos.org/jira/browse/AMBR-857

## What this PR does:
Presents the review DOI as a link to doi.org at the bottom of the review.

# Code Author Tasks:

If I used outside libraries:

# Code Reviewer Tasks

### DIFF
--- a/src/main/resources/peer-review-transform.xsl
+++ b/src/main/resources/peer-review-transform.xsl
@@ -153,8 +153,17 @@
         <xsl:apply-templates select="supplementary-material" />
       </dl>
     </xsl:if>
+
+    <xsl:variable name="review-doi-url">
+      <xsl:text>https://doi.org/10.1371/journal.</xsl:text><xsl:value-of select="../@id"/>
+    </xsl:variable>
+    <div class="review__doi">
+      <a href="{$review-doi-url}">
+        <xsl:value-of select="$review-doi-url"/>
+      </a>
+    </div>
   </xsl:template>
-  
+
   <xsl:template match="p">
     <xsl:copy>
       <xsl:apply-templates />
--- a/src/test/java/org/ambraproject/wombat/service/PeerReviewServiceImplTest.java
+++ b/src/test/java/org/ambraproject/wombat/service/PeerReviewServiceImplTest.java
@@ -31,8 +31,8 @@ import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.spy;
 
 
-@ContextConfiguration(classes = {PeerReviewServiceTest.class})
-public class PeerReviewServiceTest extends AbstractTestNGSpringContextTests {
+@ContextConfiguration(classes = {PeerReviewServiceImplTest.class})
+public class PeerReviewServiceImplTest extends AbstractTestNGSpringContextTests {
 
   private PeerReviewServiceImpl service = new PeerReviewServiceImpl();
 
--- a/src/test/java/org/ambraproject/wombat/service/PeerReviewServiceImplTest.java
+++ b/src/test/java/org/ambraproject/wombat/service/PeerReviewServiceImplTest.java
@@ -118,12 +118,9 @@ public class PeerReviewServiceImplTest extends AbstractTestNGSpringContextTests 
                 "crepoUuid", UUID.randomUUID().toString()
             ))
         )).build();
-
-
+    
     PeerReviewService serviceWithMockedContent = getServiceWithMockedContent();
-
     String html = serviceWithMockedContent.asHtml(itemTable);
-    System.out.println(html);
     Document d = Jsoup.parse(html);
 
     assertThat(d.select(".review-history .review__doi").get(0).text(), containsString("https://doi.org/10.1371/journal.pone.0207232.r001"));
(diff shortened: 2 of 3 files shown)


src/test/java/org/ambraproject/wombat/service/PeerReviewServiceImplTest.java
```diff
@@ -89,7 +89,48 @@ public void testAsHtml() throws IOException {
...
+                "crepoKey", "info:doi/10.1371/journal.pone.0207232.r001.xml",
+                "crepoUuid", UUID.randomUUID().toString()
+            ))
+        )).build();
+
+
+    PeerReviewService serviceWithMockedContent = getServiceWithMockedContent();
+
+    String html = serviceWithMockedContent.asHtml(itemTable);
+    System.out.println(html);
```
egh: Debugging statement?

src/main/resources/peer-review-transform.xsl
```diff
@@ -153,8 +153,17 @@
         <xsl:apply-templates select="supplementary-material" />
       </dl>
     </xsl:if>
+
+    <xsl:variable name="review-doi-url">
+      <xsl:text>https://doi.org/10.1371/journal.</xsl:text><xsl:value-of select="../@id"/>
+    </xsl:variable>
+    <div class="review__doi">
```
egh: Why two underscores?

src/test/java/org/ambraproject/wombat/service/PeerReviewServiceImplTest.java
```diff
@@ -89,7 +89,48 @@ public void testAsHtml() throws IOException {
...
+    Document d = Jsoup.parse(html);
+
+    assertThat(d.select(".review-history .review__doi").get(0).text(), containsString("https://doi.org/10.1371/journal.pone.0207232.r001"));
+  }
+
+  PeerReviewServiceImpl getServiceWithMockedContent() {
+    return new PeerReviewServiceImpl() {
```
egh: It's kind of a bad pattern to stub/mock/spy on the class under test. Is there some way that we can avoid this?

src/test/java/org/ambraproject/wombat/service/PeerReviewServiceImplTest.java
sgatchell: woops! thx.

src/main/resources/peer-review-transform.xsl
sgatchell: Following the convention Sam established: https://github.com/PLOS/wombat/blob/develop/src/main/resources/peer-review-transform.xsl

src/test/java/org/ambraproject/wombat/service/PeerReviewServiceImplTest.java
sgatchell: Alternatives might not come cheap. I'll timebox an attempt.

src/main/resources/peer-review-transform.xsl
egh: Excellent, ty!

src/test/java/org/ambraproject/wombat/service/PeerReviewServiceImplTest.java
sgatchell: The primary [method being mocked](https://github.com/PLOS/wombat/blob/36bb47eb2755d40d21aad46b102e8dad1a24ff98/src/main/java/org/ambraproject/wombat/service/PeerReviewServiceImpl.java#L243) is intended to encapsulate http connection and streaming methods.  So, it's existence already a change to accommodate tests, which is sometimes a smell but I think appropriate in this case.  I could extract a utility class to get content from the contentApi as a string so that the mocked collaborator is a class instead of a method.  

Mocking the individual http and streaming collaborators would be more challenging and I'm not sure about the benefit.

egh: OK, I see your point. :) Now is probably not the time to rearchitect this, but the impossibility of mocking out everything but the class under test here makes me wonder about some of the architecture underlying this. I started trying to do it, but I see your point - there is not a clear separation between the underlying HTTP code and the data model used here (`ContentKey`, etc) which is making this much more complicated that necessary. Oh well. :(

sgatchell: Well, now you've left me eager to practice the sorts of improvements you made to these tests.  :)  

It seems like a lot more test boilerplate if I were to extract a `ContentApiUtil` class, or alternatively move the logic to a `corpusContentApi.getAsString()`.  Does this sound like the correct approach to you if I needed to mock the content of a collaborating class? https://stackoverflow.com/questions/35086806/use-different-spring-test-context-configuration-for-different-test-methods.  Might be easier to just create a different test class per context.

egh: Well, feel free to give it a shot :) I tried, but I found that the `CorpusContentApi` has an unfortunate API that mixes low-level HTTP with high-level stuff like `ContentKey`. This makes it hard because you can't just mock out the HTTP stuff or the high level stuff, but both mixed together, and you have to set up the relationships between the two mocked bits properly.

(This is what I like most about tests, they really help you figure out the awkward parts of the API, because if you find yourself mocking out this much, it seems to indicate to me an awkward underlying API).

PR #1535, opened 2019-04-01
Empowerment - xsl clarity

sgatchell:
## What this PR does:

Reconciles immaterial differences in comments and whitespace between our mobile and desktop article xsl.  These diffs obscure the material differences and make it extra intimidating to make any edits.  

Post-clarification we can more easily identify differences that appear unintentional and may be bugs. We can also identify tech debt where we have chosen to make changes in xsl that could more easily be done in CSS (with features which may not have been broadly supported at the time the xsl was written).

Vast swathes of the mobile and desktop xsl are identical. Individual tags like `<citation>` have minor differences that may not be meaningful but we're afraid to touch it, partially because it's buried in immaterial differences.

I do not believe we have a means to test these changes against our corpus of articles and confirm no harm has been done, which is likely another reason we are timid about making edits.

I began this effort in anticipation of having to incorporate the article xsl (and in turn the world of plos-themes and its resource inheritance chain) into the peer review xsl as part of https://jira.plos.org/jira/browse/AMBR-864.  In my experiments the resulting display is not much better and sometimes worse and it does not bear out that we should present the same JATS tags the same way in the article pages and the peer review page.

# Code Reviewer Tasks
- [ ] PLEASE DO NOT MERGE. I would like to get broad feedback from multiple reviewers who have some history with these transforms before making this change.

### DIFF
--- a/src/main/webapp/WEB-INF/themes/desktop/xform/article-transform.xsl
+++ b/src/main/webapp/WEB-INF/themes/desktop/xform/article-transform.xsl
@@ -86,27 +86,30 @@
   <!-- Ambra modifications -->
   <xsl:template name="version-notes">
     <div class="amendment-retraction">
-    <xsl:choose>
-      <xsl:when test="front/notes[@notes-type='version-unavailable']">
+      <xsl:choose>
+        <xsl:when test="front/notes[@notes-type='version-unavailable']">
           <h2> Version no longer available.</h2>
-      </xsl:when>
-      <xsl:when test="front/notes[@notes-type='article-temporarily-unavailable']">
+        </xsl:when>
+        <xsl:when test="front/notes[@notes-type='article-temporarily-unavailable']">
           <h2> Article temporarily unavailable.</h2>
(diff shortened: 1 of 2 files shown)


src/main/webapp/WEB-INF/themes/desktop/xform/article-transform.xsl
```diff
@@ -1617,40 +1616,40 @@
...
-    <xsl:template match="list">
+  <!-- Ambra modifications -->
+  <xsl:template match="list">
+    <xsl:call-template name="newline1"/>
+    <xsl:if test="title">
+      <h5><xsl:value-of select="title"/></h5>
+    </xsl:if>
+    <xsl:choose>
+      <xsl:when test="@list-type='bullet'">
         <xsl:call-template name="newline1"/>
```
sgatchell: This whole block should be indent changes but it resulted in one line post-edit having the same indent as another line pre-edit which makes the diff look awkward.

PR #1541, opened 2019-04-12
AMBR-922 display multiple editors

sgatchell:
*JIRA issue:* https://jira.plos.org/jira/browse/AMBR-922

> Display as many editors as there are present in the sub-article xml

# Code Reviewer Tasks

### DIFF
--- a/src/main/resources/peer-review-transform.xsl
+++ b/src/main/resources/peer-review-transform.xsl
@@ -75,7 +75,7 @@
               <span itemprop="author" itemscope="" itemtype="http://schema.org/Person">
                 <span itemprop="name">
                   <!-- decision letter editor -->
-                  <xsl:apply-templates select="front-stub/contrib-group/contrib[1]" />
+                  <xsl:apply-templates select="front-stub/contrib-group" />
                 </span>
               </span>
             </div>
@@ -90,8 +90,13 @@
     </tr>
   </xsl:template>
 
-  <xsl:template match="contrib">
-    <xsl:value-of select="concat(normalize-space(./name/given-names),' ',normalize-space(./name/surname),', Editor')" />
+  <xsl:template match="contrib-group">
+    <xsl:for-each select="contrib">
+      <xsl:value-of select="concat(normalize-space(./name/given-names),' ',normalize-space(./name/surname),', Editor')" />
+      <xsl:if test="position() != last()">
+        <xsl:text>, </xsl:text>
+      </xsl:if>
+    </xsl:for-each>
   </xsl:template>
 
   <xsl:template match="sub-article[@specific-use = 'acceptance-letter']">
--- a/src/main/resources/peer-review-transform.xsl
+++ b/src/main/resources/peer-review-transform.xsl
@@ -75,7 +75,7 @@
               <span itemprop="author" itemscope="" itemtype="http://schema.org/Person">
                 <span itemprop="name">
                   <!-- decision letter editor -->
-                  <xsl:apply-templates select="front-stub/contrib-group" />
+                  <xsl:apply-templates select="front-stub/contrib-group/contrib" />
                 </span>
               </span>
             </div>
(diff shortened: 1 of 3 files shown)


src/test/java/org/ambraproject/wombat/service/PeerReviewServiceImplTest.java
```diff
@@ -228,18 +228,18 @@ public void testAttachmentAnchorTitles() throws IOException {
...
 
-    // ORIGINAL SUBMISSION 
+    // ORIGINAL SUBMISSION
 
     assertThat(doc.select(".review-history .revision .letter__title").get(0).text(), containsString("Original Submission"));
     assertThat(doc.select(".review-history .revision .letter__date").get(0).text(), containsString("June 1, 2018"));
 
     assertThat(doc.select(".review-history .decision-letter .letter__date").get(0).text(), containsString("September 12, 2018"));
-    assertThat(doc.select(".review-history .decision-letter span[itemprop=name]").get(0).text(), containsString("Qinghui Zhang, Editor"));
+    assertThat(doc.select(".review-history .decision-letter span[itemprop=name]").get(0).text(), containsString("Qinghui Zhang, Editor, Surachai Supattapone, Editor"));
```
egh: What do you think about changing this to 
```java
    assertEquals("Qinghui Zhang, Editor, Surachai Supattapone, Editor", 
                 doc.select(".review-history .decision-letter span[itemprop=name]").get(0).text());
```
so that we catch an accidental trailing `,`?

src/main/resources/peer-review-transform.xsl
```diff
@@ -90,8 +90,13 @@
     </tr>
   </xsl:template>
 
-  <xsl:template match="contrib">
-    <xsl:value-of select="concat(normalize-space(./name/given-names),' ',normalize-space(./name/surname),', Editor')" />
+  <xsl:template match="contrib-group">
+    <xsl:for-each select="contrib">
```
egh: If you like you can change this to 
```xslt
  <xsl:template match="contrib">
    <xsl:value-of select="concat(normalize-space(./name/given-names),' ',normalize-space(./name/surname),', Editor')" />
    <xsl:if test="position() != last()">
      <xsl:text>, </xsl:text>
    </xsl:if>
  </xsl:template>
```

and the `select` on line 78 would be `front-stub/contrib-group/contrib`

src/test/resources/peer-review/peer-review.pone.0207232.xml
```diff
@@ -17,6 +17,13 @@
             </name>
             <role>Academic Editor</role>
           </contrib>
+          <contrib contrib-type="author">
+            <name name-style="western">
```
egh: From a brief look at the the other XSLT it seems we don't support other name-styles. Do we know that this is the case?

src/test/java/org/ambraproject/wombat/service/PeerReviewServiceImplTest.java
sgatchell: good idea, tightening it up...

src/test/resources/peer-review/peer-review.pone.0207232.xml
sgatchell: I don't recall asking anybody but I remember coming to the conclusion with Matt after reading the explicit comments in the article transform, as you likely observed:

`<!-- commented out eastern/western logic. edit when we implement name-style eastern (possibly use generic nlm) -->`  

Do you want me to look into it further?

egh: No, if you and Matt had the same conclusion I did then :+1:

src/main/resources/peer-review-transform.xsl
sgatchell: I like!  Gotta remember that `position()` is not just for `for-each`.

PR #1565, opened 2019-08-07
AMBR-1039

msingh05:
TPR attachment link broken when accessing attachments from sub-doi

*JIRA issue:* https://jira.plos.org/jira/browse/AMBR-1039

*Related PRs:* 
> molten, plos-formulas, other applications

## What this PR does:

>Explain in a few sentences what functionality changed, and how you approached those changes. Don't be afraid to give a little extra detail. The Reviewer is going to read the original ticket, but this can point them in the right direction.

## Developer Notes

> Anything that was particularly difficult, or interesting, or
made you nervous, and should get particular attention during the review?

# Code Author Tasks:

- [ ] There are unit tests sufficient for both positive and negative test cases, test coverage has not decreased as part of this work. 

If I used outside libraries:

# Code Reviewer Tasks

### DIFF
--- a/src/main/resources/peer-review-transform.xsl
+++ b/src/main/resources/peer-review-transform.xsl
@@ -181,7 +181,7 @@
   <xsl:template match="supplementary-material">
       <dd class="supplementary-material">
         <a class="supplementary-material__label coloration-white-on-color" 
-           href="{concat('article/file?id=10.1371/journal.',@id,'&amp;type=supplementary')}"
+           href="{concat('file?id=10.1371/journal.',@id,'&amp;type=supplementary')}"
            title="{string-join(('Download',plos:get-file-extension(normalize-space(caption/p/named-content)),'file'),' ')}"
            target="_blank">
           <xsl:value-of select="label"/>
--- a/src/main/resources/peer-review-transform.xsl
+++ b/src/main/resources/peer-review-transform.xsl
@@ -181,7 +181,7 @@
   <xsl:template match="supplementary-material">
       <dd class="supplementary-material">
         <a class="supplementary-material__label coloration-white-on-color" 
-           href="{concat('file?id=10.1371/journal.',@id,'&amp;type=supplementary')}" 
+           href="{concat('article/file?id=10.1371/journal.',@id,'&amp;type=supplementary')}"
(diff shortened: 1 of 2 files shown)


src/main/resources/peer-review-transform.xsl
```diff
@@ -181,7 +181,7 @@
   <xsl:template match="supplementary-material">
       <dd class="supplementary-material">
         <a class="supplementary-material__label coloration-white-on-color" 
-           href="{concat('file?id=10.1371/journal.',@id,'&amp;type=supplementary')}" 
+           href="{concat('article/file?id=10.1371/journal.',@id,'&amp;type=supplementary')}"
```
sgatchell: Does the link still work when you start from the article page?

sgatchell:
FYI @msingh05 I was considering changing this 
https://github.com/PLOS/wombat/blob/c999210af1e9ef1283cc089743c4bcabd3554121/src/main/java/org/ambraproject/wombat/controller/PeerReviewController.java#L58
to something like `@RequestMapping(name = "peerReviewItem", value = "/article/peerReviewItem")`

Probably also needs a change here:
https://github.com/PLOS/wombat/blob/383e84b019519313cb667503290aaf5e7f53c8ec/src/main/java/org/ambraproject/wombat/controller/GeneralDoiController.java#L153

But now that I think about it that means that we'd have to change the code that registers the DOI with crossref or setup a redirect.

It's probably easier to manipulate the path like you are currently doing, and make it dependent on whether the current request includes 'article' in the path.  Anyway, just my sharing my thoughts.

msingh05:
@sgatchell I handled it through javascript. please take a look again.

src/main/webapp/WEB-INF/themes/root/resource/js/pages/peer-review.js
```diff
@@ -24,4 +24,13 @@ $(document).ready(function () {
   if(reviewEl){
     PeerReview.togglePeerReviewAccordion($(reviewEl));
   }
+
+  $('a.supplementary-material__label').each(function(index, item) {
+    var href = $(item).attr('href');
```
sgatchell: since `href` isn't used outside of the `if` you could move it in there for clarity.

PR #1572, opened 2019-09-26
AMBR-936 Remove dependency on xom

egh:
Required a switch in Base64 library.

*JIRA issue:* https://jira.plos.org/jira/browse/AMBR-936

## What this PR does:

This library was doing nothing except providing a base64 implementation, which is built in to java.

# Code Reviewer Tasks

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -164,11 +164,6 @@
       <classifier>jdk15</classifier>
     </dependency>
     <dependency>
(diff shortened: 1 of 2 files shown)


egh:
@gt-plos do you know what is going wrong with the coverage test here? I don't think this change should change the test coverage

egh:
@gt-plos I rebased against master and that fixed it :)

PR #1577, opened 2019-10-11
Ambr 1064

sgatchell:
*JIRA issue:* https://jira.plos.org/jira/browse/AMBR-1064

*Related PRs:* 
https://github.com/PLOS/plos-themes/pull/1000

## What this PR does:
Updates the mobile themes to use journal-branded amendments unless specifically overridden, as is the case for retraction and eoc which are red and correction which is off-white.  This brings the amendment view to parity with desktop.

Also added the updated-article relationship type to the set of relationships that deserve and get a blurb at the top center of the mobile page

## Developer Notes

Very annoying that the desktop themes use SASS while the mobile just use CSS.  I do not understand it and I don't know how complex it would be to reconcile the desktop and mobile themes.

### DIFF
--- a/src/main/webapp/WEB-INF/themes/mobile/resource/css/interface.css
+++ b/src/main/webapp/WEB-INF/themes/mobile/resource/css/interface.css
@@ -222,6 +222,8 @@
 
 /* Alerts */
 .red-alert,
+.amendment-retraction,
+.amendment-eoc,
 .amendment-uncorrected-proof {
   background: #fce2e5;
 }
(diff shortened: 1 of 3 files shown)


src/main/webapp/WEB-INF/themes/mobile/resource/css/interface.css
```diff
@@ -240,6 +242,10 @@
   margin-top: 0 !important;
 }
 
+.amendment-correction {
```
sgatchell: All amendments currently have a red background on mobile, in contrast with desktop.  This brings mobile to parity with desktop where corrections are off-white, retractions and EOCs are red, and others are a journal-branded shade

sgatchell: I went a little rogue and changed corrections to off-white for parity with desktop. They were previously red like retractions. Just checking if I should have kept them red in mobile? @samV-plos

PR #1587, opened 2020-01-07
Merge plos-themes code into wombat

egh:


wryfi:
@egh as we discussed last quarter, Ann made a business decision that we should not merge the plos-themes code into wombat if we are to keep wombat open source, viewing the look and feel of our site as proprietary trade dress that we do not want to enable other parties to easily copy.  Please let me know if you have any questions.

PR #1591, opened 2020-01-22
AMBR-1069 Show preceding article citation in updating articles

egh:
Fix for https://jira.plos.org/jira/browse/AMBR-1069?focusedCommentId=271219&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-271219

image

### DIFF
--- a/src/main/webapp/WEB-INF/themes/desktop/ftl/article/amendmentGroupTitle.ftl
+++ b/src/main/webapp/WEB-INF/themes/desktop/ftl/article/amendmentGroupTitle.ftl
@@ -20,4 +20,9 @@
   ~ DEALINGS IN THE SOFTWARE.
   -->
 
-${amendmentGroup.type.displayName}
+<#if amendmentGroup.type.name == "updated-article">
+  Preceding Article
+<#else>
+  ${amendmentGroup.type.displayName}
+</#if>
+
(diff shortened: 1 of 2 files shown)


src/main/webapp/WEB-INF/themes/desktop/ftl/article/amendmentGroupTitle.ftl
```diff
@@ -20,4 +20,9 @@
   ~ DEALINGS IN THE SOFTWARE.
   -->
 
-${amendmentGroup.type.displayName}
+<#if amendmentGroup.type.name == "updated-article">
```
sgatchell: Cool, and plosone overrides this with it's own behavior https://github.com/PLOS/plos-themes/blob/develop/code/all_devices/journals/PlosOne/ftl/article/amendmentGroupTitle.ftl

It works but it's fragile.  Good thing this template and the overriding one customize the same amendment group type, otherwise some logic might be accidentally obliterated for the overriding journal.

egh: Yes, I don't really like this. I considered changing it [here](https://github.com/PLOS/wombat/blob/develop/src/main/java/org/ambraproject/wombat/model/RelatedArticleType.java#L163) but the display on the sidebar ("UPDATE pertains to") is different from the citation. :(

sgatchell: Well, it was a good thought.  For now we use plos-themes as intended.

PR #1592, opened 2020-01-22
AMBR-1103 stop calling nonexistent redirect logging service

sgatchell:
http://journals.vagrant.test/plosone/redirect/guidelines#fungal
Screenshot from 2020-01-22 15-37-03
You should no longer see a 403 error in the Network tab of your browser tools when you invoke a client-side redirect. You will, however, get a 404 because there is no static page service configured in vagrant

compare to:
https://journals-dev.plos.org/plosone/redirect/guidelines#fungal
Screenshot from 2020-01-22 15-29-18

### DIFF
--- a/src/main/webapp/WEB-INF/themes/root/ftl/redirect.ftl
+++ b/src/main/webapp/WEB-INF/themes/root/ftl/redirect.ftl
@@ -44,15 +44,6 @@ This page has moved. Please click <a href="<@siteLink path='${defaultTarget}'/>"
     }
   }
 
-  // Ajax URL should be set as a redirect in Apache. The redirect destination is arbitrary and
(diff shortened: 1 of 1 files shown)


egh:
I don't see a 403 error when accessing  https://journals.plos.org/plosone/redirect/guidelines#fungal - am I missing something?

Nonetheless, this seems pointless so we may as well get rid of it :+1:

noxmwalsh:
Screen Shot 2020-01-23 at 10 24 25 AM

@egh - I'm seeing the 403 when checking out the same link in Chrome

egh:
It was in Firefox, but I don't see it in Chrome either. Must be an ad blocker or something. In any case, it doesn't really matter.

PR #1596, opened 2020-02-18
AMBR-977: Implement updated Figshare widget

msingh05:


sgatchell:
@msingh05 how do I test this?

src/main/webapp/WEB-INF/themes/root/resource/js/components/figshare.js
```diff
@@ -47,4 +71,28 @@ if (window.figshare) {
     window.loadedWidgets = loadedWidgets;
 
   });
+
+
+  // doubleclickad gets added to <aside> element of this widget. so rename it to <div>
```
sgatchell: I want to understand this problem better.  What's an example article that I can use to see the issue?

msingh05:
> @msingh05 how do I test this?

Figshare widget should work as same, there are some minor visual changes.

src/main/webapp/WEB-INF/themes/root/resource/js/components/figshare.js
msingh05: you can reproduce the issue with any article which has supporting information. Comment the function call and you can see the issue. I am attaching the screenshot also.

msingh05: Screen Shot 2020-02-19 at 3 05 11 PM

sgatchell:
I observed a bug in the new widget.  The bottom description does not change when you select a different file.  See how it's showing the description of the the first file even though I've scrolled to 3/3.  The dev environment does not have this bug - it updates the description to match the selected file.
Screenshot from 2020-02-20 08-52-54

msingh05:
@sgatchell can you check again. marketing has fixed their code so I removed the hack.

PR #1601, opened 2020-03-05
AMBR-1121 Number of ONE Academic Editors Out of Date on Website

msingh05:


src/main/webapp/WEB-INF/themes/desktop/sass/edboard.scss
```diff
@@ -8,7 +8,7 @@
 }
 
 a {
-  color: blue;
+  color: #3e0577;
```
gt-plos: Can we use $plos-one-purple?

PR #1603, opened 2020-03-10
AMBR-1119 Env var configuration

egh:
Changing wombat to use env variables for all runtime configuration.

New env vars to use:
- `RHINO_URL`: complete url to rhino server
- `SOLR_URL`: complete url to solr server, including collection
- `CAS_URL`: base url for CAS server
- `NED_URL`: the URL for the ned server, including username ans password
- `MECACHED_SERVER`: (optional) `hostname:port` for memcached server to use
- `LOG_LEVEL`: level of logging, defaults to `warn`
- `ROOT_PAGE_PATH`: the path to the root page to use
- `THEME_PATH`: path to the themes directory 

Note that in the future, when we finish dockerizing this, if the themes are embedded in the docker container, as I think they will be, we will not need to use those last two env vars and they can be hardcoded.

Other changes:
- all logs now go to stdout
- no longer uses slf4j, only log4j

Example of usage:
```
SOLR_URL=http://solr-mega-dev.soma.plos.org:80/solr/journals_dev RHINO_URL=http://journals-dev1-backend1.soma.plos.org:8006/v2/ CAS_URL=https://nedcas-integration.plos.org/cas ROOT_PAGE_PATH=/home/egh/c/ambra/var/themes/content/rootPage.html THEME_PATH=/home/egh/c/ambra/var/themes/ NED_URL=http://dipro:XXXXX@nac-390.soma.plos.org:8888/v1/ mvn package cargo:run
```


src/main/java/org/ambraproject/wombat/controller/CommentController.java
```diff
@@ -78,11 +78,6 @@
   @Autowired
   private RuntimeConfiguration runtimeConfiguration;
 
-
-  private void addCommentAvailability(Model model) {
-    model.addAttribute("areCommentsDisabled", runtimeConfiguration.areCommentsDisabled());
-  }
```
JackLaBarba: Do you know when this setting might have been used in the past? I wonder if it was a quick way to turn off commenting if somebody was massively spamming us. If that's a concern, do you think removing this as a config value would hinder our ability to respond to those situations? (sorry, there's a lot of hypotheticals there)

sgatchell: I was told that the intent was to have an easy switch to disable commenting for whatever reason.  I am unaware that it was ever used but I can see the value.  Might be better to keep the feature and hardcode the config value.

src/main/java/org/ambraproject/wombat/service/remote/UserApiImpl.java
```diff
@@ -53,16 +54,14 @@
...
-        this.server = new URL(Objects.requireNonNull(server));
+        this.server = new URL(server.getScheme(), server.getHost(), server.getPort(), server.getPath());
       } catch (MalformedURLException e) {
         throw new RuntimeException(e);
       }
-
-      this.authorizationHeader = ((authorizationAppName != null) && (authorizationPassword != null))
-          ? ImmutableList.of(createAuthorizationHeader(authorizationAppName, authorizationPassword))
-          : ImmutableList.of();
+      String[] userInfo = server.getUserInfo().split(":");
```
JackLaBarba: I wonder if it would be better to do this at bean init time (in `setNedUrl`), to surface missing user strings ASAP

src/main/java/org/ambraproject/wombat/controller/CommentController.java
JackLaBarba: If we want to keep it, it should probably be Env var configurable. In the types of scenarios where I imagine we'd want to turn it off (scenarios where some abuse is happening), we'd want to do it in a hurry, and cutting a release to do so would be burdensome.

sgatchell: That's a good point.  I'm wary of removing the feature.  Perhaps @shanecyr could weigh in?

egh: I'm happy to revert this and make it an optional config with a sensible default (false). I may have gotten overzealous. But I'll wait to hear from Shane.

src/main/java/org/ambraproject/wombat/service/remote/UserApiImpl.java
egh: Yes, it probably would be.

egh: 00183d387

src/main/java/org/ambraproject/wombat/controller/CommentController.java
JackLaBarba: We determined in standup that this feature can burn in hell. Closing out this thread . . .

JackLaBarba:
👍 ok! I think this looks good @egh, I'm going to go ahead and merge it.

PR #1619, opened 2020-04-15
Eng 55

mjvaldez:
This PR adds a jib configuration for creating a docker image for wombat. I've also added a docker-compose.yml for testing.

I considered a few ways to implement themes. For now I've chosen to just mount these as a volume. The reason being it's ultimately the most flexible way to include data, including in a k8s deployment, where the data could be attached as a persistent volume claim that abstracts the actual data source type (e.g. could be GCS).

To build the docker image, run `mvn clean package jib:dockerBuild` or you can pull these:
- plos/rhino:2.5.9-SNAPSHOT
- plos/wombat:3.7.17-SNAPSHOT

First set the necessary environment variables (explained in docker-compose.yml). Then docker-compose up, wait a minute and open (http://localhost:8015/wombat/DesktopGeneric). You should see "You have setup Wombat correctly!"

docker-compose will start a mysql instance on port 3306 so stop anything listening on that port. This port can(should?) be changed to something else for local development but I'm not sure if we have a standard for that yet.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,12 +1,12 @@
 version: '3'
 
-# NOTE: You must set WOMBAT_TAG, RHINO_TAG and PLOS_THEMES_REPO environment variables
+# NOTE: You must set PLOS_THEMES_REPO environment variables
 #       as well as all environment variables not explicitly defined below (e.g. CAS_URL)
 
 services:
 
   wombat:
-    image: plos/wombat:${WOMBAT_TAG}
+    image: plos/wombat:latest
     depends_on:
       - db
       - rhino
@@ -27,7 +27,7 @@ services:
       - ${PLOS_THEMES_REPO}:/opt/plos/themes:ro
 
   rhino:
-    image: plos/rhino:${RHINO_TAG}
+    image: plos/rhino:latest
     depends_on:
       - db
     environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,52 @@
+version: '3'
+
+# NOTE: You must set WOMBAT_TAG, RHINO_TAG and PLOS_THEMES_REPO environment variables
+#       as well as all environment variables not explicitly defined below (e.g. CAS_URL)
+
+services:
+
+  wombat:
+    image: plos/wombat:${WOMBAT_TAG}
+    depends_on:
+      - db
+      - rhino
+    environment:
+      - CAS_URL
+      - NED_URL
+      - SOLR_URL
+      - ENVIRONMENT=dev
+      - LOG_LEVEL=debug
+      - RHINO_URL=http://rhino:8080/rhino/
+      - THEME_PATH=/opt/plos/themes
+    links:
+      - db
+      - rhino
+    ports:
+      - 8015:8080
+    volumes:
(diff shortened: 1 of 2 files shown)


egh:
@mjvaldez is this ready for review?

egh:
image

pom.xml
```diff
@@ -556,7 +556,25 @@
...
+      <plugin>
+        <groupId>com.google.cloud.tools</groupId>
+        <artifactId>jib-maven-plugin</artifactId>
+        <version>2.1.0</version>
+        <configuration>
+          <to>
+            <image>docker.io/plos/wombat:${project.version}</image>
+          </to>
+          <from>
+            <image>docker.io/plos/tomcat:latest</image>
```
egh, line 569: docker complains about this but I assume because we control the image it's OK

egh:
FYI @mjvaldez because I don't think an `ENVIRONMENT` configuration is very compatible with 12-factor, and because it's not really explicit and is used only for showing debug info, I'm planning to modify it here: https://github.com/PLOS/wombat/pull/1620

pom.xml
egh: docker complains because the build is not reproducible when using `latest`

docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+      - LOG_LEVEL=debug
+      - RHINO_URL=http://rhino:8080/rhino/
+      - THEME_PATH=/opt/plos/themes
+    links:
+      - db
+      - rhino
+    ports:
+      - 8015:8080
+    volumes:
+      - ${PLOS_THEMES_REPO}:/opt/plos/themes:ro
```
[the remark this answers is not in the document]
mjvaldez: I could go either way but I figured a volume was most flexible. I'll create another ticket to address that specifically.

mjvaldez:
Makes sense. Will there be any new env vars from that change?

docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+
+  rhino:
+    image: plos/rhino:${RHINO_TAG}
+    depends_on:
+      - db
+    environment:
+      - AWS_ACCESS_KEY_ID
```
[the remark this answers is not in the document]
mjvaldez: Derp. Will remove.

pom.xml
mjvaldez: That's technically true with any tag (or by having RUN statements in your Docekrfile) but yes I think it's ok in this case. We can version at some point if it's a concern.

egh:
Yes, there will be a new env var but it won't be required.

docker-compose.yml
```diff
@@ -0,0 +1,48 @@
+version: '3'
+
+# NOTE: You must set WOMBAT_TAG, RHINO_TAG and PLOS_THEMES_REPO environment variables
+#       as well as all environment variables not explicitly defined below (e.g. CAS_URL)
+
+services:
+
+  wombat:
+    image: plos/wombat:${WOMBAT_TAG}
```
egh, line 9: since docker-compose will only be used for local dev (right?) what do you think about just using `latest` here? (and below for rhino)

mjvaldez: fair point and removes the need for that var. updated.

PR #1624, opened 2020-05-05
Eng 119

egh:
https://jira.plos.org/jira/browse/ENG-119

~~Written against the docker branch because that is easier to configure debug logging (to track the number of solr queries) and also because that is the future.~~

Rebased on develop.

Changes:
- Major rewrite of Solr code to ensure that result data is in proper java objects, rather than a lot of untyped code. I also refactored code around to be in more (to me) logical places.
- After this, I was able to identify a few places where solr queries code be reduced:
  1. The faceted queries could be reduced, because multiple facets can be returned from a single query
  1. The desktop search did 2x the searches required, because it uses a dynamic search which threw away the first results returned by the wombat server

Now each search should require only 3 searches, not 10. If memcache is used, 1 of these queries should be loaded from the cache, so there are really only 2 queries made.

The other currently required query is a sub-document faceting query which could probably also be cleaned up.

As far as I can tell, almost all of the time waiting for search results is spent waiting for solr now.

To test:

Run
```
$ mvn package cargo:run
```

Or similar. You should enable the memcached server because the taxonomy browser relies on solr faceting and uses memcached to cache its results, so if you don't have a memcached server you will get poor performance.

~~You should see 3 solr searches for every request to search. ~~

If you want to see how many queries it makes, you'll need to figure out how to enable debug logging or add a `System.out.println` statement to the `SolrSearchApi` class.


benjaminkreen:
NICE! I'm seeing a drop from 12 requests down to 3! Thats amazing. No comment on the code, thats a way bigger java diff than i'm comfortable with.

src/main/java/org/ambraproject/wombat/controller/SearchController.java
```diff
@@ -652,59 +645,12 @@ private ModelAndView getFeedModelAndView(Site site, String feedType, String titl
...
-   *
-   * @param request HttpServletRequest
-   * @param model   model that will contain search results
-   * @param site    site the request originates from
-   * @param params  all URL parameters
-   * @return String indicating template location
-   * @throws IOException
-   */
-
-  @RequestMapping(name = "csvExport", value = "/csvExport", produces = MediaType.APPLICATION_OCTET_STREAM_VALUE)
```
sgatchell, line 674: :feelsgood: getting rid of unused features.  I see later that this whole controller is remove, extra nice!

egh: I should have mentioned that I talked to Kristen about this & she gave me the go-ahead to get rid of it.

src/main/java/org/ambraproject/wombat/service/remote/SolrQueryBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+
+    if (asq.getFacetFields().size() > 0) {
+      params.add("facet", "true");
+      for (String field: asq.getFacetFields()) {
+        params.add("facet.field", field);
+      }
+      params.add("facet.mincount", Integer.toString(asq.getFacetMinCount()));
+      params.add("facet.limit", Integer.toString(asq.getFacetLimit()));
+      /* Use a thread per facet. */
+      params.add("facet.threads", Integer.toString(-1));
```
sgatchell, line 73: Thanks for explaining the magic number! :thread: :thread: :thread:

sgatchell:
browsing subject areas is slow.  Do you think it's due to caching or the change in code?
image

egh:
> browsing subject areas is slow. Do you think it's due to caching or the change in code?

Almost certainly caching. If you are able you can install memcached and configure it in your `wombat.yaml`. You should see faster browsing then.

src/main/java/org/ambraproject/wombat/service/remote/SolrQueryBuilder.java
egh: Makes me think I should use a constant....

sgatchell:
ah, much better with memcached, thank you

src/main/java/org/ambraproject/wombat/service/BrowseTaxonomyService.java
```diff
@@ -41,20 +40,18 @@
    * For the current journal return a complete structured map of the taxonomic categories
    *
    * @param journalKey the current journal
-   * @param site the current site
    * @return a complete structured map of the taxonomic categories
    */
-  TaxonomyGraph parseCategories(String journalKey, Site site) throws IOException;
```
sgatchell, line 47: glad to see that this can go

src/main/java/org/ambraproject/wombat/service/SearchFilterService.java
```diff
@@ -73,87 +74,36 @@
...
-        .setQuery(query.getQuery().orElse(null))
-        .setSimple(query.isSimple())
-        .setDateRange(query.getDateRange().orElse(null))
-        .setJournalKeys(query.getJournalKeys())
-        .setSubjects(query.getSubjects())
-        .setAuthors(query.getAuthors());
-
-    Map<?, ?> sectionFacetResults = solrSearchApi.search(sectionFacetQuery.build(), site);
+  public Map<String, SearchFilter> getSearchFilters(ArticleSearchQuery query) throws IOException {
+    ImmutableList<String> facetFields = ImmutableList.of(JOURNAL_FACET_FIELD,
```
sgatchell, line 84: Just noting that the inconsistent use of the `_FIELD` suffix annoys me.

src/main/java/org/ambraproject/wombat/service/remote/SolrQueryBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+
+    if (asq.isPartialSearch()) {
+      params.add("qf", "doc_partial_body");
+      params.add("fl", "*");
+      params.add("fq", "doc_type:partial");
+    } else {
+      params.add("fq", "doc_type:full");
+    }
+
+    params.add("fq", "!article_type_facet:\"Issue Image\"");
```
sgatchell, line 48: We do this for saved searches too.  When is this facet wanted I wonder? I also wonder if solr could be configured to skip this facet by default.

egh: Perhaps we should stop indexing issue images? If we always want to exclude them, the simplest thing would be to ensure they aren't indexed :)

egh: They are still indexed http://solr-mega-prod.soma.plos.org/solr/journals_prod_20191101/select?fq=article_type_facet:%22Issue%20Image%22&q=*:*

src/main/java/org/ambraproject/wombat/service/SearchFilterService.java
egh: fixed in 5e795c06d

src/main/java/org/ambraproject/wombat/service/remote/SolrQueryBuilder.java
egh: 12d2d50e2

PR #1626, opened 2020-05-06
Empowerment: load resources more efficiently

benjaminkreen:
Just something thats been bugging me regarding website performance

Regarding image tags: modern browsers will only load these once they are nearing the viewport. When a user lands on a desktop article page nearly all images are loaded offscreen anyways. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Attributes

Regarding script tags: I only added defer to third party standalone stuff, so it shouldn't affect much. (crossref is just a button at the top, ~~figshare is a widget at the bottom~~ and mathjax mucks around with formula formats) See https://alligator.io/html/defer-async/
EDIT: undid the figshare defer, there is some synchronous code that depends on it. Thinking of a better solution.

Thoughts?
Our most viewed articles of the year have abysmal load times:
image

For DOI 10.1371/journal.pone.0083152 with warm cache it looks like this:

OLD
image
NEW
image

Slimmer pages!

### DIFF
--- a/src/main/webapp/WEB-INF/themes/desktop/ftl/article/aside/crossmark.ftl
+++ b/src/main/webapp/WEB-INF/themes/desktop/ftl/article/aside/crossmark.ftl
@@ -23,7 +23,7 @@
  <@themeConfig map="article" value="showCrossMarkWidget" ; showCrossMarkWidget>
     <#if showCrossMarkWidget>
 <!-- Crossmark 2.0 widget -->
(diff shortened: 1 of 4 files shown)


egh:
This is very cool, but I feel like it deserves to be a ticket and get proper QA. I'm not sure what the implications of this are.

benjaminkreen:
c'mon, live a little ;) jk, i'll look into that

benjaminkreen:
closed in favor of #1627

PR #1630, opened 2020-05-28
ENG-119 Fix problem with journal, article type faceting

egh:
https://jira.plos.org/jira/browse/ENG-119

Fix for problem #2 identified by @gt-plos  here https://jira.plos.org/jira/browse/ENG-119?focusedCommentId=281957&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-281957

Explanation: Article type and journal are facets that should be ORed together, not ANDed together. That is, when a user is attempting to facet, they want articles in ONE **OR** Medicine, not both in ONE **AND** Medicine (which is impossible). In order to support this, we need to exclude certain facet queries from the faceting mechanism. That is, we want to return facet counts as though the facet query did not happen, because we want to return the counts that a user would get if they added an OR clause
to the facet query. Fortunately, solr supports this out of the box using tagging/excluding filters.

For more details see:
https://lucene.apache.org/solr/guide/7_4/faceting.html#Faceting-TaggingandExcludingFilters

To test, compare dev (broken)
https://journals-dev.plos.org/plosone/search?filterJournals=PLoSCompBiol&filterSubjects=Neurons&filterArticleTypes=Review&q=&page=1
image

with prod (working) https://journals.plos.org/plosone/search?filterJournals=PLoSCompBiol&filterSubjects=Neurons&filterArticleTypes=Review&q=&page=1

image

Notice that even with restrictions on which journals and article types, we still have options available. Compare that with dev, where there are no options available.

And with this fix, those options are again restored:
http://localhost:8080/DesktopPlosOne/search?filterJournals=PLoSCompBiol&filterSubjects=Neurons&filterArticleTypes=Review&filterArticleTypes=Editorial&q=&page=1
image

### DIFF
--- a/src/test/java/org/ambraproject/wombat/service/remote/SolrSearchApiTest.java
+++ b/src/test/java/org/ambraproject/wombat/service/remote/SolrSearchApiTest.java
@@ -119,7 +119,7 @@ public class SolrSearchApiTest extends AbstractJUnit4SpringContextTests {
 
   private List<NameValuePair> buildFacetParams(String facetField, String query, boolean useDisMax) {
     ArticleSearchQuery asq = ArticleSearchQuery.builder()
-      .setFacetFields(ImmutableList.of(facetField))
+      .addFacet(facetField)
       .setQuery(query)
       .setSimple(useDisMax)
       .build();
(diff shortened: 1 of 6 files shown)


sgatchell:
Tangentially I'm thinking this pain should be managed in one place by a service that translates PLOS search idioms into a solr query. This should be a problem we're solving separately in python for our saved searches, right?

egh:
> This should be a problem we're solving separately in python for our saved searches, right?

I guess so, I hadn't thought of it like that. However this itself should only effect the facet results, which don't matter for saved searches. But there probably is some duplicated logic in saved searches. I'm not sure how to eliminate that shared logic in a sensible way, given that one codebase is python and one is java.

src/test/java/org/ambraproject/wombat/service/remote/SolrSearchApiTest.java
```diff
@@ -143,6 +143,20 @@ public void testFacetParams() {
...
+      .setField("foo")
+      .setExcludeKey("bar")
+      .build();
+    ArticleSearchQuery asq = ArticleSearchQuery.builder()
+      .addFacet(facet)
+      .setQuery("query")
+      .build();
+    actual = SolrQueryBuilder.buildParameters(asq);
+    actualMap = convertToMap(actual);
+    System.out.println(actualMap.get("facet.field"));
```
sgatchell, line 157: logging

egh: Oops, fixed in 
5d0ea3c

PR #1633, opened 2020-06-04
ENG-211

egh:
https://jira.plos.org/jira/browse/ENG-211

This was a journey. What I learned:

1. We do not use memcached for session storage, but instead use tomcat's built in mechanism.
2. Session storage for spring actually seems to happen in tomcat (the container). Our best bet for not using sticky sessions any more is probably tomcat’s “cluster” https://tomcat.apache.org/tomcat-8.5-doc/cluster-howto.html option.
3. A lot about docker and docker-compose

In the end, I found there seems to be no obvious problem with sharing memcached between tomcats.

I have made some changes to the docker build and `docker-compose.yml` to allow testing this.

1. Added memcached to the `docker-compose.yml`
2. Added a dead-simple load balancer to docker to allow us to scale wombat horizontally.
3. Updated the README to how to make docker-compose work.

To test:
1. Follow new README instructions (https://github.com/plos/wombat/blob/ENG-211/README.md#docker)
2. Add `RHINO_URL=http://journals-dev1-backend1.soma.plos.org:8006/v2/` to your `.env` to use existing rhino content.
3. Run `docker-compose up --scale wombat=3` to start with 3 wombats.
4. Visit http://localhost:8080/DesktopPlosOne/article/comments?id=10.1371/journal.pone.0205192 in multiple browser windows, or choose some other way of hammering on the server to try to find any issues.

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -1,0 +1,17 @@
+user  nginx;
+
+events {
+    worker_connections   1000;
+}
+
+http {
+    server {
+        listen 8080;
+        location / {
+            # Docker uses round robin DNS, so this will load balance
+            # between the wombat instances. Note that this will not
+            # work with sessions.
+            proxy_pass http://wombat:8080;
+        }
+    }
+}
--- a/README.md
+++ b/README.md
@@ -14,8 +14,9 @@ files, as well as containing several configuration options.
 Runtime configuration happens via environment variables:
 
 - `CAS_URL`: base url for CAS server
-- `LOG_LEVEL`: (optional) level of logging, defaults to `warn`
+- `COLLECTIONS_URL`: URL for the collections server, used for links
 - `DEBUG`: (optional) set to true to show debug info if a page errors
+- `LOG_LEVEL`: (optional) level of logging, defaults to `warn`
 - `MECACHED_SERVER`: (optional) `hostname:port` for memcached server to use
 - `NED_URL`: the URL for the ned server, including username and password
 - `RHINO_URL`: complete url to rhino server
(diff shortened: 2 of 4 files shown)


nginx.conf
```diff
@@ -0,0 +1,14 @@
...
+
+events {
+    worker_connections   1000;
+}
+
+http {
+    server {
+        listen 8080;
+        location / {
+            proxy_pass http://wombat:8080;
```
sgatchell, line 11: I don't understand where the load balancing logic resides. The upstream is a black box from nginx's perspective, is that right? I also didn't see that `docker-compose up --scale` implies load balancing.

mjvaldez:
If you want to avoid the volume mount, you could merge the themes-submod branch which includes the themes as a submodule.

sgatchell:
We'd love to, but PLOS designed themes to be a private extension of the wombat public OSS offering, which is the main reason they haven't been merged together. Moving to gitlab is an opportunity to rethink this old strategy, which is leaky anyway since there is PLOS-specific content in wombat.

nginx.conf
egh: Ah, good point! I should add an explanatory note. Docker compose uses round robin dns, so `wombat` should resolve to each of the wombat instances.

egh:
> We'd love to, but PLOS designed themes to be a private extension of the wombat public OSS offering, which is the main reason they haven't been merged together. Moving to gitlab is an opportunity to rethink this old strategy, which is leaky anyway since there is PLOS-specific content in wombat.

I think the submodule method should be safe, because the submodule handles authn distinctly.

egh:
> cool demo of compose scaling

PR #1635, opened 2020-06-12
ENG-123 Bugsnag for wombat

egh:
https://jira.plos.org/jira/browse/ENG-123

This is mostly straightforward. Fortunately bugsnag support works out of the box with our version of spring (unlike PLOS Match).

- This is configured by two new env vars, `BUGSNAG_API_KEY` and `BUGSNAG_RELEASE_STAGE`. Both are optional. If `BUGSNAG_API_KEY` is not provided, no bugsnag reporting will occur. If `BUGSNAG_RELEASE_STAGE` is not set, the default of `production` is used.
- All uncaught exceptions will now be reported to bugsnag if an API key is provided.
- In places where `log.error(...)` was logging an exception, I added a `bugsnag.notify(...)`
- I eliminated the `RuntimeConfiguration` interface. In my opinion, an interface with a single implementation is an anti-pattern that increases the amount of code in the app for no benefit. I also eliminated a test class that could be replaced by a mock.

## Testing
1. Visit bugsnag.com and select the wombat project, then go to project settings. Find the API key
image
2. Start wombat locally
```
$ BUGSNAG_RELEASE_STAGE=development \
BUGSNAG_API_KEY=XXXX \
RHINO_URL=http://journals-dev1-backend1.soma.plos.org:8006/v2/ \
SOLR_URL=http://solr-mega-dev.soma.plos.org:80/solr/journals_dev/ \
CAS_URL=https://nedcas-integration.plos.org/cas \
NED_URL=http://dipro:XXX@nac-390.soma.plos.org:8888/v1/ \
COLLECTIONS_URL=http://collections-dev.plos.org \
THEME_PATH=/path/to/plos-themes mvn package cargo:run
```
3. Visit http://localhost:8080/DesktopPlosOne/ and make the site error. One easy thing you can do is not connect to the VPN, this should generate an error.
4. Visit bugsnag.com and verify that the error was reported there.


src/main/java/org/ambraproject/wombat/config/RuntimeConfiguration.java
```diff
@@ -22,58 +22,145 @@
...
 
-
 /**
- * Interface that represents configurable values that are only known at server startup time.
+ * Class that represents configurable values that are only known at server
+ * startup time.
  */
-public interface RuntimeConfiguration {
-
+public class RuntimeConfiguration {
```
sgatchell, line 38: Agreed, the interface is clutter unless you need it for Spring DI.

PR #1638, opened 2020-06-22
ENG-195 solr configs and README

wryfi:
This PR imports the `journals_v1_0_3` configuration from solr-mega-prod, which is the configuration currently in use by the production journals stack.  If some other configuration would be better suited to check in here, please let me know.  

It also adds a `README` with instructions for managing solr configs via docker.

Is there anything else that should go in here?

### DIFF
--- a/solr/README.md
+++ b/solr/README.md
@@ -3,6 +3,14 @@
 This folder contains the solr schema and configuration for the PLOS journals
 stack. 
 
+In SOMA, solr is deployed and managed via 
+[solrcloud-formula](https://github.com/PLOS-formulas/solrcloud-formula).
+
+In GCP, solr is deployed and managed via kubernetes and deployed via the 
+[solr](https://gitlab.com/plos/gcp-global/-/tree/master/modules/solr) 
+terraform module in 
+[gcp-global](https://gitlab.com/plos/gcp-global).
+
 # Cloud Configuration 
 
 ## Overview
--- a/solr/README.md
+++ b/solr/README.md
@@ -1,0 +1,112 @@
+# solr 
+
+This folder contains the solr schema and configuration for the PLOS journals
+stack. 
+
+# Cloud Configuration 
+
+## Overview
+
+A solr configuration consists of a named hierarchy of files. Each collection on
+a cluster must use a configuration that has been uploaded to solr. Many
+configurations can be stored, but only one can be used per collection.
+
+The configurations for solr cloud are stored in zookeeper. The easiest way to
+manipulate them is to use the `zkcli.sh` script bundled with solr. (Note that solr's 
+`zkcli.sh` is NOT the same as the `zkCli.sh` that comes bundled with zookeeper.) 
+The easiest way to run `zkcli.sh` is via `docker run`.
+
+For example ... 
+
+```
+docker run --rm -v ~/repos/plos/wombat/solr:/solrconf solr:7.7 \
+./server/scripts/cloud-scripts/zkcli.sh -z solr-zoo-201.soma.plos.org:2181/solr-mega \
+-cmd upconfig -confdir /solrconf/config -confname wombat_31d72ab_2020-06-22
+```
+
+You can also download configurations and perform a handful of other useful
+tasks with `zkcli.sh`. To see all the available options:
+
+```
+docker run --rm solr:7.7 ./server/scripts/cloud-scripts/zkcli.sh help
+```
+
+If you don't want to use docker for some reason, you can download the solr
+release and run `zkcli.sh` from there. 
+
+
+## Naming Conventions
+
+When uploading a configuration to solr, name it after the repo and hash where
+the configuration can be found in version control, along with the date of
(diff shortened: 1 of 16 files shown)


solr/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+
+## Naming Conventions
+
+When uploading a configuration to solr, name it after the repo and hash where
+the configuration can be found in version control, along with the date of
+creation. This will make it easy to see at glance which configuration is
+newest, while also providing an easy reference to the source code where the
+config can be located. For configurations deployed from this repo, the name in
+solr should generally look something like `wombat_31d72ab_2020-06-22`.
```
JackLaBarba, line 53: Can we move the date in front of the hash? That way alphanumeric sort gives us a chronological sort

wryfi: Good call @JackLaBarba, updated accordingly.

wryfi:
Updated to remove some PLOS implementation details.

solr/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+In this mode, a single node hosts a single core and there is no concept of
+collections. The configuration directory is provided as an argument to the
+`create_core` command.
+
+To run a standalone solr core with this configuration on docker:
+
+```
+docker run --rm -d --name localsolr -v ~/repos/plos/wombat/solr:/solrconf -p 8983:8983 solr:7.7
+docker exec -it --user=solr localsolr solr create_core -c journals -d /solrconf/config
```
egh, line 76: I'm getting this error:
```
🤔 docker exec -it --user=solr localsolr solr create_core -c journals -d /solrconf/config

ERROR: Error CREATEing SolrCore 'journals': Unable to create core [journals] Caused by: com.cominvent.solr.RequestSanitizerComponent
```

We seem to use some sort of request sanitizer plugin which hasn't been updated in 3 years. We use it to set a maximum value for rows of 1000. It's not clear to me if we still need this, but if we do I think we need to copy in the jar to our docker instance.

wryfi: @egh yeah our SOMA configuration uses some solr sanitizer plugin to prevent people from taking down the cluster with a huge query (see RARO-233).  So I had to remove the plugin-related configurations to use the container, because the jar isn't deployed or configured there.  

Not sure what to do about this for SOMA.  I've been working (in my spare time) on [solrizr](https://gitlab.com/wryfi/solrizr) which I was hoping we could use in GCP.  I guess we could also deploy it in SOMA.

Another solution would be to roll our own docker image with the sanitizer plugin installed and enabled.

solr/config/synonyms.txt
```diff
@@ -0,0 +1,208 @@
...
+Somalis => Somalian people
+Swedes => Swedish people
+Bugis => Buginese people
+Oriyas => Odishan people
+Sasak => Sasak people
+Pardo Brazilians => Pardo Brazilian people
+Obsessive�compulsive personality disorder => Obsessive-compulsive personality disorder
+Georgia (United States) => Georgia (United States state)
+Linguistic morphology => Morphology (linguistics)
+Archaeans => Archaea
```
JackLaBarba, line 208: the dev and stage journals collections actually use a synonyms.txt that's slightly expanded from the current prod one. See: http://solr-mega-dev.soma.plos.org/solr/#/journals_stage/files?file=synonyms.txt

Note 

The reason for this is because the dev and stage synonyms.txt contain entries for the 2017-2 thesaurus release, but 2017-2 was never deployed on prod (intentionally anyways, the 2017-2 snuck its way into prod which cause this bug: https://jira.plos.org/jira/browse/AMBR-1085).

solr/README.md
egh: I was able to fix this issue locally by downloading the request sanitizer jar, changing the solr config slightly and volume mounting it into docker. See the following:

```patch
diff --git a/solr/README.md b/solr/README.md
index c91322929..077be9248 100644
--- a/solr/README.md
+++ b/solr/README.md
@@ -72,6 +72,7 @@ collections. The configuration directory is provided as an argument to the
 To run a standalone solr core with this configuration on docker:
 
 ```
+wget https://github.com/cominvent/request-sanitizer-component/releases/download/v1.1/request-sanitizer-1.1.0.jar -P solr/lib/
 docker run --rm -d --name localsolr -v ~/repos/plos/wombat/solr:/solrconf -p 8983:8983 solr:7.7
 docker exec -it --user=solr localsolr solr create_core -c journals -d /solrconf/config
 ```
diff --git a/solr/config/solrconfig.xml b/solr/config/solrconfig.xml
index a09ac9960..3b003d2c4 100644
--- a/solr/config/solrconfig.xml
+++ b/solr/config/solrconfig.xml
@@ -37,6 +37,7 @@
 
     <lib dir="./lib" />
     -->
+  <lib dir="/solrconf/lib/"/>
 
   <!-- 
     Used to specify an alternate directory to hold all index data
diff --git a/solr/lib/.keep b/solr/lib/.keep
new file mode 100644
index 000000000..e69de29bb
```
I am not sure how to do the same in GKE.

